PR #1, opened 2025-11-12
Add benchmarks

karlowich:
This adds a wrapper for SIL, arguments for GDSIO and two benchmarks: synthetic and file.

Potentially, we can add BaM `nvm-block-bench` to the synthetic benchmark

### DIFF
--- a/scripts/sil.py
+++ b/scripts/sil.py
@@ -1,0 +1,60 @@
+"""
+A SIL-wrapper invoking it with arguments
+========================================
+
+Retargetable: True
+------------------
+"""
+
+from argparse import ArgumentParser, _StoreAction
+
+
+def add_args(parser: ArgumentParser):
+    class StringToBoolAction(_StoreAction):
+        def __call__(self, parser, namespace, values, option_string=None):
+            setattr(namespace, self.dest, values == "true")
+
+    parser.add_argument("--backends", type=str, default=["posix"], nargs="+")
+    parser.add_argument("--dataset", type=str, default="imagenetish")
+    parser.add_argument("--device", type=str, default="/dev/nvme1n1")
+    parser.add_argument("--bin", type=str, default="sil")
+    parser.add_argument("--batch_size", type=int, default=1)
+    parser.add_argument("--batches", type=int, default=1)
+    parser.add_argument("--gpu_nqueues", type=int, default=6)
+    parser.add_argument("--repetitions", type=int, default=5)
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


scripts/sil.py
```diff
@@ -0,0 +1,48 @@
+"""
+A SIL-wrapper invoking it with arguments
+===============================================================
```
naddinadja: Shorten this line to match the header :)

tasks/benchmark_synthetic.yaml
```diff
@@ -0,0 +1,40 @@
+doc: |
+  Run benchmarks to evaluate synthetic performance with GDS and AiSIO
+  ========================================================================
```
naddinadja: Shorten this line to match the header :)

tasks/benchmark_synthetic.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+  This assumes you already ran the task 'setup_aisio.yaml'
+  
+
+steps:
+- name: benchmark_aisio_seq
+  uses: sil
+  with:
+    backends: ["aisio"]
+    dataset: [""]
```
naddinadja: Why `dataset: [""]` (also on line 25)? 
1. In your SIL wrapper, the `dataset` argument is not a list , see line 18:
    `parser.add_argument("--dataset", type=str, default="imagenetish")`
2. Will an empty string make the argument in the command `--data-dir  `? (like with empty space after it)

naddinadja (changes requested):
I would probably merge the first two commits, "add SIL wrapper" and "add args to SIL wrapper", as I don't see it necessary to have a commit with `TODO`'s and then have a fix for that immediately after :)

tasks/benchmark_synthetic.yaml
karlowich: Right, nice catch.
It needs a bit more logic I see.

scripts/sil.py
karlowich: Fixed

tasks/benchmark_synthetic.yaml
karlowich: Fixed

tasks/benchmark_synthetic.yaml
karlowich: Fixed

naddinadja (approved):
nice! lgtm :)

safl (approved):
Great! LGTM.

PR #2, opened 2025-12-09
CPU benchmarks with bdevperf

naddinadja:
Easy-access workflows! Fast benchmarks! Beautiful graphs! Buy now, get one for free! 💯 

Here you have the cijoe workflow and belonging scripts and configs to run bdevperf with a multitude of different parameters and generate the HTML that visualises it using ChartJS.

There is also a commit that expands the "Background" section of the documentation. There might be added more, and other sections will also be added to with the method+results of these benchmarks, but I wanted to create a PR with the benchmarking workflow now to get it _out there_.


naddinadja:
Example HTML page with visualisation of benchmark results:
[benchmark-results.html](https://github.com/user-attachments/files/24054109/benchmark-results.html)

safl (changes requested):
SO good!

Can you fix the conflict, then its good-to-go.

naddinadja:
done, @safl :)

PR #3, opened 2025-12-11
feat: add uio-dmabuf

karlowich:
This PR adds a separate folder for the uio-dmabuf work.
It has two examples, a patch for the linux kernel, and a README describing how to install the custom kernel and run the examples.

Additionally, the instructions in the README has been added as CIJOE task.

Currently, this is **NOT** required for the rest of the AiSIO stack, thus the separate task and folder.


safl (approved):
LGMT.

PR #6, opened 2025-12-18
feat(dmabuf):patch UDMABUF instead of UIO

karlowich:
The functionality is exactly the same, except now it doesn't require a device bound to `uio_pci_generic`


safl (approved):
Excellent!

PR #12, opened 2026-01-13
Updates to CPU benchmarks

naddinadja:
I found a bug in my code, which prompted me to make some more improvements to the workflow. All aid me in getting more correct results, and eases the process of getting and analyzing the results.


naddinadja:
Commit `87b28e6` adds numbers to the graph like this:
image

karlowich (approved):
Very nice

PR #14, opened 2026-01-21
fix(setup): add parallel to provisioning

karlowich:
GNU parallel is required for setting up the datasets. Make sure it is installed.


safl (approved):
Great, thanks!

PR #16, opened 2026-01-27
Improvements to the CPU benchmarks

naddinadja:
Different stuffs I have noticed when running experiments again and analysing the results.


scripts/bdevperf_populate.py
```diff
@@ -0,0 +1,78 @@
...
+
+def main(args, cijoe: Cijoe):
+    """Precondition block devices using bdevperf"""
+
+    devices: list = cijoe.getconf("devices")
+
+    spdk_path = cijoe.getconf("spdk.repository.path", None)
+    if not spdk_path:
+        log.error("Failed: Missing SPDK repository path in config")
+        return
```
safl: Missing return-value here?

tasks/populate_devices.yaml
```diff
@@ -0,0 +1,14 @@
...
+  Warning: This will overwrite everything on the devices defined
+  in the config file.
+
+steps:
+- name: bind_to_userspace
+  run: |
+    {{ config.xnvme.driver.prefix }} xnvme-driver
+
+- name: precondition
+  uses: bdevperf_precondition
```
safl: Should this reference "bdevperf_populate"?

docs/src/experiments.md
```diff
@@ -115,28 +115,19 @@ how other processes in the system impact the achievable IOPS.
 
 ### Execution of the Experiment
 
-The setup and execution of the experiment is automated by two cijoe workflows,
-which necessitates a cijoe configuration file. See ``configs/bench-amd.toml`` for
-an example configuration file.
+The execution of the experiment is automated by a cijoe workflow, which
+necessitates multiple cijoe configuration file. Most are given in this repository,
```
safl: Needs a plural "configuration files", right?

docs/src/experiments.md
```diff
@@ -160,23 +151,17 @@ It's important that if the boot device is also an NVMe device that it is added i
...
-
-Lastly, this experiment requires SPDK and xNVMe to be installed on the system. The
-example configuration file does not need any changes for the ``spdk.repository``
-and ``xnvme.repository`` keys.
-
 #### Setup of the experiment
 
-The cijoe workflow in ``tasks/setup_benchmark_bdevperf.yaml`` is responsible for
-installing all dependencies on the system. When the configuration file is created,
+The machine must be provisioned as describes in the AiSIO repository README file,
```
safl: This should probably be "as described"?

configs/devices_16.toml
```diff
@@ -1,23 +1,9 @@
...
-
 [xnvme.driver]
 prefix = "PCI_BLACKLIST=0000:01:00.0"
 
-[xnvme.repository]
-remote = "https://github.com/xnvme/xnvme"
-tag = "v0.7.5"
-path = "/root/git/xnvme"
-
 [[devices]]
```
safl, line 7: Perhaps rename this file to something more descriptive that conveys that this has 16 devices, then another config could be made for a single device.

scripts/bdevperf_populate.py
```diff
@@ -0,0 +1,78 @@
...
+import logging as log
+
+from cijoe.core.command import Cijoe
+
+from cpu_freq_helper import CpuFrequencyHelper
+
+
+def populate_device(cijoe: Cijoe, spdk_path: Path, pcie_addr: str) -> int:
+    fio_cmd = (
+        f"LD_PRELOAD={Path(spdk_path) / "build" / "fio" / "spdk_nvme"} "
```
safl: This nested use of ``"`` is problematic for even recent versions of Python. It would be better to either move it out of the format string or use single-quotes on the format-string.

scripts/bdevperf_populate.py
```diff
@@ -0,0 +1,78 @@
...
+    cfm = CpuFrequencyHelper(cijoe)
+
+    err = cfm.toggle_smt(True)
+    if err:
+        log.error(f"Failed: cfm.toggle_smt(true)")
+        return err
+
+    err = cfm.toggle_turbo(True)
+    if err:
+        log.error(f"Failed: cfm.toggle_turbo(true)")
```
safl: This does not need the f-string?

scripts/bdevperf_populate.py
```diff
@@ -0,0 +1,78 @@
...
+        return err
+
+    for device in devices:
+        err = populate_device(cijoe, spdk_path, device["pci_addr"])
+        if err:
+            log.error(f"Failed: populate_device({device["pci_addr"]}); err({err})")
```
safl: Here is the nested ``"`` again. Preferable to use single-quotes on the format-string when using ``"`` inside.

safl (changes requested):
A couple of things, see inline.

configs/devices_16.toml
naddinadja: What would you think about calling it "example_devices.toml"? I explain in the white paper that you can add/remove devices from this file as necessary. However, if i just called it "16_devices" (for example), one might assume that it would work on any system with 16 block devices, which is not the case.

naddinadja:
Thanks for the good catches @safl :)

configs/devices_16.toml
safl: I was thinking something as trivial as ``devices_16.toml``, ``devices_4.toml``, and ``devices_1.toml``. From the top of my head then those are the instances we are currently using.

naddinadja: okay, i named it ``devices_16.toml``, and renamed the existing ``bench-amd.toml`` to ``devices_12.toml``. I described in the docs that these are both examples and must be changed to fit the system the benchmarks are being run on. :)

Edit: scratch that, I completely removed the config for bench-amd. We are not running on that anymore, and from an outsider's perspective, it might confuse more than necessary to have multiple configuration files. The docs clearly describe how to read and modify the one example we have. :)

karlowich: Perhaps we don't need two example configs? Looks like there is no significant difference between them.

docs/src/experiments.md
```diff
@@ -194,17 +172,19 @@ When the "run" step has been parameterized, the workflow can be run with command
 
     cijoe --monitor \
       tasks/bench_bdevperf_cpu.yaml \
-      -c config/YOUR_CONFIG_FILE.toml
+      -c configs/aisio.toml \
+      -c configs/devices.toml
```
karlowich: This doesn't exist anymore right?

docs/src/experiments.md
```diff
@@ -194,17 +172,19 @@ When the "run" step has been parameterized, the workflow can be run with command
...
 repeatedly if running the benchmarks multiple times in a row. These steps can be
 skipped by specifying which steps to run in the command
 
-    cijoe --monitor \
+    cijoe \
       tasks/bench_bdevperf_cpu.yaml \
-      -c config/YOUR_CONFIG_FILE.toml \
-      run visualize
+      -c configs/aisio.toml \
+      -c configs/devices.toml
```
karlowich: Same here

docs/src/experiments.md
```diff
@@ -154,7 +154,15 @@ making the machine unusable (until next reboot).
...
-The machine must be provisioned as described in the AiSIO repository README file.
+The machine must be provisioned as described in the AiSIO repository README file,
+and the devices on the machine can be populated with the task defined in
+``tasks/populate_devices.yaml`. When the configuration file is created,
+it can be run with command
+
+    cijoe --monitor \
+      tasks/populate_devices.yaml \
+      -c configs/aisio.toml \
+      -c configs/devices.toml
```
karlowich: Same here

karlowich (changes requested):
Overall very nice, few comments

safl (approved):
lgtm

karlowich (approved):

PR #20, opened 2026-02-03
fix(config): pin SIL and XAL to tags

karlowich:
This makes it possible to develop the projects without disrupting AiSIO


naddinadja (approved):
great!

safl (approved):
lgtm

PR #26, opened 2026-03-05
refactor(homi): rename homi to homid

naddinadja:
This is as preparation for separating the homi daemon namespace from the homi client namespace. As the homi client will be included in other projects, the "homi" namespace will be used for that.


safl (approved):
lgtm.

PR #27, opened 2026-03-05
Basic interface for the HOMI IPC

naddinadja:
These commits implement the skeleton for doing IPC with HOMI. A shared protocol is defined by `homi_proto`. On the daemon side, it creates a socket and listens to that. A client library is defined by `homic`.

Because HOMI still has no real functionality, I've added just a single example endpoint "helloworld". It acts as an example of how to define input and output for communication, and how to handle it on both ends. As we add functionality to HOMI, we might want to remove this.

I've written a simple test C program using the helloworld endpoint:
[homi-ipc-test.c](https://github.com/user-attachments/files/25930320/homi-ipc-test.c)

### DIFF
--- a/homi/include/homi_proto.h
+++ b/homi/include/homi_proto.h
@@ -1,0 +1,52 @@
+#ifndef HOMI_PROTO_H
+#define HOMI_PROTO_H
+
+#include <stdint.h>
+
+#define HOMI_MAX_CONNECTS   8
+
+enum homi_msg_type {
+	/* TODO: add message types as daemon functionality is extended */
+	HOMI_MSG_TYPE_HELLOWORLD = 0, ///< Test type, as an example of what is needed
+};
+
+struct homi_req_helloworld {
+	int32_t value;
+};
+
+struct homi_msg_header {
+	enum homi_msg_type type;
+	size_t payload_len;
+};
+
+/**
+ * Read a message from a socket.
+ *
+ * Reads a homi_msg_header followed by its payload from sock_fd. The payload
+ * is heap-allocated and returned via *buf; the caller is responsible for
+ * freeing it. *buf is set to NULL if payload_len is zero.
+ *
+ * @param sock_fd  File descriptor of the connected socket.
+ * @param hdr      Output: populated with the received message header.
+ * @param buf      Output: allocated buffer containing the payload, or NULL.
+ * @return         0 on success, negative errno on failure.
+ */
+int
+homi_proto_socket_read(int sock_fd, struct homi_msg_header *hdr, char **buf);
+
+/**
+ * Write a message to a socket.
+ *
+ * Sends hdr followed by buf as a single framed message. Sets hdr->payload_len
+ * to buf_len before writing.
+ *
+ * @param sock_fd   File descriptor of the connected socket.
+ * @param hdr       Message header; payload_len will be overwritten with buf_len.
+ * @param buf       Payload to send.
+ * @param buf_len   Length of the payload in bytes.
+ * @return          0 on success, negative errno on failure.
+ */
+int
+homi_proto_socket_write(int sock_fd, struct homi_msg_header *hdr, void *buf, size_t buf_len);
+
+#endif /* HOMI_PROTO_H */
--- a/homi/src/homic.c
+++ b/homi/src/homic.c
@@ -1,0 +1,109 @@
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/un.h>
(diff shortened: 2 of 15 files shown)


homi/include/homi_proto.h
```diff
@@ -0,0 +1,53 @@
+#ifndef HOMI_PROTO_H
+#define HOMI_PROTO_H
+
+#include <stdint.h>
+
+#define HOMI_SOCKETFILE  "/run/homi/homi.sock"
```
safl: Hmm... this probably needs to be parameterized.

homi/src/homid_ipc.c
```diff
@@ -0,0 +1,166 @@
...
+		homid_log(LOG_ERR, "Failed: bind(); err(%d)", errno);
+		close(fd);
+		return -errno;
+	}
+
+	return fd;
+}
+
+int
+homid_ipc_open(struct homid_ipc_connection **conn)
```
safl: This would benefit from some input-argument with options like the location of the socketfile.

homi/src/homic.c
```diff
@@ -0,0 +1,109 @@
...
+	char *response;
+	int err;
+
+	if (!g_homic_client) {
+		err = -ENOTCONN;
+		fprintf(stderr, "Failed: No connection, please call homic_client_connect(); err(%d)\n", err);
+		return err;
+	}
+
+	req.value = (int32_t)value;
```
safl: Would it not be better that this function took a value of the proper type type instead of casting here?

homi/src/homic.c
```diff
@@ -0,0 +1,109 @@
...
+	hdr.type = (uint32_t)msg_type;
+	hdr.payload_len = (uint32_t)sizeof(req);
```
safl: Same for this.

homi/src/homic.c
```diff
@@ -0,0 +1,109 @@
...
+	int err;
+
+	if (!g_homic_client) {
+		err = -ENOTCONN;
+		fprintf(stderr, "Failed: No connection, please call homic_client_connect(); err(%d)\n", err);
+		return err;
+	}
+
+	req.value = (int32_t)value;
+	hdr.type = (uint32_t)msg_type;
```
safl: And perhaps ``hdr.type`` should be the enum?

homi/src/homid.c
```diff
@@ -43,22 +47,48 @@ parse_args(int argc, char *argv[], struct homid_cli_args *args)
 }
 
 static int
-initialize(struct homid_opts *opts)
+homid_initialize(struct homid **homid, struct homid_opts *opts)
```
safl: It is often nice to go by the convention of input first then out arguments.

safl (commented):
A couple of questions/comments.

homi/include/homi_proto.h
naddinadja: This is needed by both the daemon and client, and therefore must be defined at compile time, when both are installed. I can add it to meson instead of here, but it cannot be added as a parameter to the daemon cli. The path I use here is just the systemd `Service` argument `RuntimeDirectory=homi`, which automatically creates the "/run/homi" directory.
What benefits are gained by parameterizing it?

homi/src/homid_ipc.c
```diff
@@ -0,0 +1,166 @@
...
+		break;
+	}
+	default:
+		homid_log(LOG_WARNING, "Unknown message type: %u", hdr.type);
+		break;
+	}
+
+exit:
+	free(payload);
+	close(sock_fd);
```
siu-jung, line 137: Not really an issue but, would it be better if we close the fd in the same function we opened it?

homi/include/homi_proto.h
safl: I can't follow why this must be a compile-time constant, i must be missing something.

If this was parametrized e.g. a config in ``homid``, and and argument in the client, then one can:

* Run ``homid`` in other locations... in case that is needed e.g. when the system it runs on follows a different location convention
* Run multiple instances of the homi daemon and from the client side choose which to connect to

naddinadja: I didn't know this was neither common nor wanted, but I see! I've changed it to be a parameter in the homid configuration file :)

homi/src/homid_ipc.c
naddinadja: I can't do that, because it's inside a worker thread. So the main thread just spawns a worker, let's it do the work and close the `sock_fd`, so that the main thread can continue accepting requests

siu-jung: Oh yeah sorry, my bad

safl (approved):
lgtm

PR #29, opened 2026-03-11
refactor(udmabuf-import): move to external repo

karlowich:
Now that udmabuf-import is available at github.com/xnvme/udmabuf-import we don't need a copy in AiSIO.


safl (approved):
lgtm

PR #32, opened 2026-03-18
feat(homi): add configs, linkage, init for xnvme and xallib

siu-jung:
Parse xal options: backend, watchmode, file lookupmode.
Added files src/homid_xal.c and include/homid_xal.h

Assumes user cijoe'ed prior to building homi
Also added 'make status' for convenience

Setup xnvme and xal and init structures
Use default xnvme opts with linux backend

### DIFF
--- a/homi/include/homid.h
+++ b/homi/include/homid.h
@@ -1,3 +1,5 @@
 struct homid {
   struct homid_ipc_connection *conn;
+  struct homid_device *dev;	///< Pointer to array of 'struct homid_device'
+  unsigned int ndevs;		///< Number of devices
 };
--- a/homi/include/homid_xal.h
+++ b/homi/include/homid_xal.h
@@ -1,0 +1,57 @@
+struct homid_opts;
+
+struct homid_device {
+	struct xnvme_dev *dev;
+	struct xal *xal;
+};
+
+/**
+ * Setup xal for the homid_device
+ *
+ * For the given homid_device, open xal and retrieve extents through a full scan.
+ * xnvme device must be initialized first.
+ *
+ * @param opts		xal_opts parsed from config file.
+ * @param device	Output: device to setup.
+ * @return			0 on success, negative errno on failure.
+ */
+int
+homid_xal_setup(struct xal_opts *opts, struct homid_device *device);
+
+/**
+ * Setup xnvme for the homid_device
+ *
+ * For the given homid_device, initialize xnvme.
+ * Uses default xnvme_opts with "linux" as backend.
+ *
+ * @param uri		URI of the device.
+ * @param device	Output: device to setup.
+ * @return			0 on success, negative errno on failure.
+ */
+int
+homid_xnvme_setup(char *uri, struct xnvme_dev **device);
+
+/**
+ * Cleans up array of homid_device
+ *
+ * For the given number of devices, clean each homid_device's xal and xvnme setups.
+ * Checks before freeing.
+ *
+ * @param ndevs		Number of devices to clean.
+ * @param devices	Array of homid_device to clean.
+ */
+void
+homid_device_close(unsigned int ndevs, struct homid_device *devices);
+
+/**
+ * Setup array of homid_device
+ *
+ * Allocates array of homid_device and initializes xnvme and xal for each of them.
+ * Uses device uri, ndevs, and xal_opts from user config options.
+ *
+ * @param opts		config options parsed from user config file.
+ * @param devices	Output: array of homid_device to setup.
+ * @return			0 on success, negative errno on failure.
+ */
+int
+homid_device_setup(struct homid_opts *opts, struct homid_device **devices);
--- a/homi/src/homid.c
+++ b/homi/src/homid.c
@@ -7,11 +7,13 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <libxal.h>
+
 #include <homid.h>
 #include <homid_ipc.h>
 #include <homid_log.h>
+#include <homid_xal.h>
 #include <homid_opts.h>
-
 
 volatile sig_atomic_t stop = 0;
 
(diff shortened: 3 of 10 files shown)


homi/src/homid_xal.c
```diff
@@ -0,0 +1,123 @@
...
+#include <homid_xal.h>
+#include <homid_opts.h>
+
+int
+homid_xal_setup(struct homid_device *device, struct xal_opts *opts)
+{
+	struct xal *xal;
+	int err;
+
+	err = xal_open(device->dev, &xal, opts);
```
naddinadja, line 25: You probably need a check on whether device exists such that it doesn't break if `!device` and then you try to check `device->dev`

homi/include/homid.h
```diff
@@ -1,3 +1,4 @@
 struct homid {
   struct homid_ipc_connection *conn;
+  struct homid_device *dev;
```
naddinadja: I would add a docstring for this, potentially:
```///< Pointer to array of 'struct homid_device'```
and then also add a `uint32_t ndevs;  ///< Number of of devices` here, so that the array is accompanied with its size. :)

File: homi/include/homid_xal.h [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+homid_xal_setup(struct homid_device *dev, struct xal_opts *opts);
+
+int
+homid_xnvme_setup(struct homid_device *device, char *uri);
+
+void
+homid_device_close(struct homid_device *devices, int ndevs);
+
+int
+homid_device_setup(struct homid_opts *opts, struct homid_device **devices);
```
naddinadja, line 57: Can we add documentation strings to these? :)

homi/src/homid.c
```diff
@@ -138,6 +141,7 @@ int main(int argc, char **argv)
 	homid_log(LOG_NOTICE, "Daemon terminated");
 
 exit:
+	homid_device_close(homid->dev, opts.ndevs);
```
naddinadja: Since the setup of the devices is inside the `homid_initialize()` funciton, I would expect this to be in the `homid_close()`. If we add the `ndevs` field to the homid struct, we can call that instead of `opts.ndevs`. :)

homi/src/homid_opts.c
```diff
@@ -105,6 +109,57 @@ homid_opts_from_toml(char *config_file, struct homid_opts *opts)
...
+	if (xal_backend.u.int64 == XAL_BACKEND_XFS)
+		opts->xal_opts.be = XAL_BACKEND_XFS;
+	else if (xal_backend.u.int64 == XAL_BACKEND_FIEMAP)
+		opts->xal_opts.be = XAL_BACKEND_FIEMAP;
+	else {
+		homid_log(LOG_ERR, "Missing or invalid 'xal.backend' property in config");
+		err = -EINVAL;
+		goto exit;
+	}
+
```
naddinadja, line 128: This is just a style thing, but we always use curly brackets around `if` and `for` statements, even if they're just a single line. I see this a couple other places, but I'll just comment one time here :)

homi/include/homid_xal.h
```diff
@@ -0,0 +1,18 @@
...
+struct homid_device {
+	struct xnvme_dev *dev;
+	struct xal *xal;
+};
+
+int
+homid_xal_setup(struct homid_device *dev, struct xal_opts *opts);
+
+int
+homid_xnvme_setup(struct homid_device *device, char *uri);
```
naddinadja: Can we change this to `int homid_xnvme_setup(char *uri, struct xnvme_dev **xnvme_dev)`, where `xnvme_dev` is the output pointer? It's a small simplification, but it helps keeping the homid_device logic in the setup function, and only having the xnvme logic in this function :)

naddinadja (changes requested):
Thank you so much for this! I have a couple of comments inline, but overall it looks great! :)

homi/include/homid.h
siu-jung: I added it as unsigned int, if that's ok? :)

homi/include/homid_xal.h
siu-jung: done!

homi/include/homid_xal.h
siu-jung: Fixed!
Also changed all the other functions so that argument ordering is input -> output.

homi/src/homid_xal.c
siu-jung: Done!

homi/src/homid.c
siu-jung: Yeah you're right. Fixed it!

homi/src/homid_opts.c
siu-jung: Ok haha added them all!

homi/src/homid_xal.c
```diff
@@ -0,0 +1,127 @@
...
+		if (err) {
+			homid_log(LOG_ERR, "Failed to setup XAL for %s: %d", uri, err);
+			goto failed;
+		}
+	}
+
+failed:
+	*devices = devs;
+	return err;
+}
```
naddinadja, line 130: I would move the `*devices = devs;` to above the "failed:", and return 0, and then inside the "failed", you could just free(devs). This way, if something goes wrong, it doesn't exit the function :)

naddinadja (changes requested):
I have one more thing here :)

homi/include/homid.h
naddinadja: yea, that's probably fine too :)

homi/src/homid_xal.c
siu-jung: Ok will do that! But a single free(devs) wouldn't be enough to clean inits from the previous for loop iterations (if there are multiple devices). I added a call to homid_device_close() instead!

naddinadja (approved):
This looks good to me now :) @safl, do you want to give it a look too?

safl (approved):
lgtm!

PR #33, opened 2026-03-18
Docs: sharpen taxonomy, challenges, and add authors

safl:
The work spans a wide area in different directions, getting the framing right is essential to clearly communicate why and what we are doing. Arguably, the this reframing carves it out more clearly.

### DIFF
--- a/docs/src/architecture.md
+++ b/docs/src/architecture.md
@@ -27,160 +27,200 @@ contemporary operating systems, Linux uniquely satisfies these requirements at
 scale, and therefore serves as the reference platform for the architectures and
 implementations described in this work.
 
-Modern systems require such coexistence not because workloads favor different
-initiators, but because interoperability with operating system managed storage
-is a hard requirement. Applications, filesystems, and OS infrastructure depend
-on host-initiated I/O paths that provide safety, consistency, and universal
-abstractions. These paths cannot simply be replaced. At the same time, emerging
-workloads motivate specialized fast paths that reduce host involvement and
-enable direct device-to-device data movement.
+AiSIO designs adopt a structural separation between control-oriented
+responsibilities and high-bandwidth data-path execution. Metadata management,
+protection enforcement, and system coordination remain associated with the
+operating system, while accelerators participate in data-path execution using
+mechanisms suited to parallel data movement. This separation allows each
+processing unit to operate in its area of strength, while preserving the safety
+and correctness guarantees that applications and file systems depend on.
 
-Many AiSIO designs therefore adopt a structural separation between
-control-oriented responsibilities and high-bandwidth data-path execution.
-Metadata management, protection enforcement, and system coordination remain
-associated with the operating system, while accelerators participate in
-data-path execution using mechanisms suited to parallel data movement. This
-separation allows each processing unit to operate in its area of strength, while
-preserving the safety and correctness guarantees expected from operating system
-managed storage.
+## I/O Path Taxonomy
 
-This architecture is particularly relevant for AI workloads where GPUs process
-massive datasets stored in filesystems, but the underlying requirements extend
-well beyond AI. Any scenario that demands coexistence between OS-managed storage
-and high-performance direct access benefits from multipath I/O. Scientific
-computing, database acceleration, video processing, and network-attached storage
-exhibit similar pressures on the storage stack.
+To reason about how AiSIO system architectures support multipath coexistence, we
+characterize I/O paths along three independent axes: infrastructure, initiator,
+and buffer placement.
 
-To reason about how such system software architectures support this form of
-coexistence in practice, we distinguish between three classes of I/O paths
-commonly encountered in AiSIO systems.
+**Infrastructure** describes which software layer owns the NVMe driver and
+manages the controller's queues.
 
-**OS-managed** I/O paths provide the interoperability baseline. They preserve
-the conventional kernel storage stack, including filesystems, VFS layers,
-permission enforcement, and existing applications. All file operations proceed
-through standard interfaces, ensuring correct metadata handling, consistency,
-and durability. This path maintains the safety and semantic guarantees required
-by the operating system and its applications.
+- *Kernel-managed*: the kernel NVMe driver performs controller initialization,
+  queue allocation, command submission, and completion handling. The kernel
+  retains full control over device management, file-system integration, and
+  protection enforcement.
+- *User-space managed*: a user-space NVMe driver performs these responsibilities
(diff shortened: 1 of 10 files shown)


docs/src/architecture.md
```diff
@@ -27,170 +27,212 @@ contemporary operating systems, Linux uniquely satisfies these requirements at
...
+| Components                        | Infrastructure | Initiator | Buffers |
+| --------------------------------- | -------------- | --------- | ------- |
+| pread, libaio, io_uring           | Kernel         | CPU       | Host    |
+| SPDK                              | User-space     | CPU       | Host    |
+| xNVMe/uPCIe                      | User-space     | CPU       | Host    |
+| GDS                               | Kernel         | CPU       | Device  |
+| io_uring + dma-buf                | Kernel         | CPU       | Device  |
+| xNVMe/uPCIe + CUDA/dma-buf       | User-space     | CPU       | Device  |
+| xNVMe/uPCIe + ROCm/dma-buf       | User-space     | CPU       | Device  |
+| xNVMe/uPCIe + CUDA (device-resident) | User-space | Device    | Device  |
```
karlowich: I think we should consider using either CPU/GPU or Host/Device. That is, perhaps we should change to host-initiated instead of CPU-initiated.

docs/src/architecture.md
```diff
@@ -27,170 +27,212 @@ contemporary operating systems, Linux uniquely satisfies these requirements at
...
+  accelerator must access them directly.
+
+These axes are independent. The following table summarizes how existing and
+AiSIO system architectures map onto the three axes:
+
+| Components                        | Infrastructure | Initiator | Buffers |
+| --------------------------------- | -------------- | --------- | ------- |
+| pread, libaio, io_uring           | Kernel         | CPU       | Host    |
+| SPDK                              | User-space     | CPU       | Host    |
+| xNVMe/uPCIe                      | User-space     | CPU       | Host    |
```
karlowich: How much coverage is intended? Should we add libvfn here too?

karlowich (changes requested):
Two things for now

docs/src/architecture.md
safl: Yeah... I did mull on this for a while... when saying "host-initiated" then I get a sense that it is not sufficiently clear. E.g. it seems like it lags a qualifier like "host-cpu-initiated" and then like "device-processing-unit-initiated". When just labelling then "CPU-initiated" and "device-initiated" then the assumption is that the CPU in question is on the host, and although the device might actually have a processing unit which could also be a  CPU, but it just happens to be in a peripherial device. This separation is sort of an implicit assumption that the architectural context addressed here is host-cpu with device connected via pcie.
And I guess there are different areas to explore even when mentioning this. E.g. the potential efficiency of CPUs vs GPUs (as the mainstream massively parallel processors from NVIDIA and AMD) and the challenges of discrete accelerator similar to the PCIe attached GPUs from AMD and NVIDIA in contrast to both integrated GPUs on the same die as the CPU as well as GPUs interconnected via C2C links like NVLINK with the actual CPU.

This rambling is a bit unclear. Let me try to rephrase. When saying "host-initiated" then that becomes unclear when considering that it is common for a host to have a main processing-unit that takes care of the operating system context and that it consists of a CPU and GPU integrated on the same die, e.g. AMD APUs, Intel iGPU, NVIDIAs superchip etc. So, considering these, does host-initiated also cover initiating I/O from the these GPU or from the CPU on the processing-unit in the host? Here prefer the phrase ``CPU-initiated`` since the focus is on the ability of the x86/arm like main processing units ability to initiate I/O and not of the GPU in this case.
I then prefer the ``device-initiated`` since it sets the expectation of a peripheral device as one challenge and area of interest there is the management of ``P2P`` memory and physically separated memory in addition to handling I/O initiation from a processing unit with a significantly different instruction-set and architecture.

This was probably not super clear, so very briefly, yes I do prefer using the terms of "CPU-initiated" and "Device-initiated".

docs/src/architecture.md
```diff
@@ -27,170 +27,212 @@ contemporary operating systems, Linux uniquely satisfies these requirements at
...
+
+**Infrastructure** describes which software layer owns the NVMe driver and
+manages the controller's queues.
+
+- *Kernel-managed*: the kernel NVMe driver performs controller initialization,
+  queue allocation, command submission, and completion handling. The kernel
+  retains full control over device management, file-system integration, and
+  protection enforcement.
+- *User-space managed*: a user-space NVMe driver performs these responsibilities
+  directly from user space. This eliminates kernel transitions, interrupt
```
karlowich: User-space is spelled in two different ways within the same sentence

docs/src/architecture.md
safl: I wanted a scope that aligns to what is described in the introduction. Then the reader has a mental anchor and can compare that to the components introduced with the work done under the AiSIO initiative.

safl: E.g. it can help the reader understand the separations when they might already be familiar with kernel-based approaches, SPDK, and GDS. At least these a touched upon in the description, so it could be primed in their mind when reading this.

safl: For that mentioning libnvm/BaM/SCADA here might be good as well.

karlowich: Sounds good.

docs/src/architecture.md
karlowich: In that context (considering host setups with CPU/GPU mixed), is it clear that host memory is CPU accessible DRAM? Or do we face the same problem on the memory side?
The phrasing as questions is intentional, I can't make up my mind.

safl: Host with tightly integrated CPU/GPU **often** share the same memory, that is, both processing units access the same physical memory. As the main attraction for the integration is unifying the memory. E.g. for Intel iGPUs, AMD APUs, NVIDIA GB10 as superchip example, and even Apple Unified memory on their Apple silicon.

karlowich: All good then

docs/src/architecture.md
naddinadja: let's do a mega-haul of user space spelling after this PR, such that we fix it in every place :)

safl: I went ahead and addressed it along with a couple of other related ones. However.... overstreched (broke labels0 so fixing that now.

safl: fixed.

karlowich (approved):
LGTM

naddinadja (approved):
LGTM :)

PR #34, opened 2026-03-18
docs: fix results table and comment syntax

safl:
Split compound table cells into separate columns to remove inline HTML <br> tags. Convert link reference data annotations to MyST comments using the % prefix.


naddinadja (approved):
thanks! :)

PR #36, opened 2026-03-30
fix(configs/udmabuf): change repos. path to '/root/git'

safl:
Changing this for two reasons:

* Align with the paths in the other configs assume that repositories live in '/root/git'.

* When using the task with a remote system then it is likely that 'local.env.home' does not properly reflect a path on the remote system, especially not when the remote system is setup as described in "setup_ubuntu".


karlowich (approved):
LGTM

I must have seen `local.env.HOME` used somewhere else (probably xNVMe), that's why I added it originally.

PR #37, opened 2026-03-31
Individual benchmark tool scripts

naddinadja:
I've added individual scripts for the three benchmarking tools bdevperf, spdk_nvme_perf and xnvmeperf, such that these can be used inside our larger benchmarking scripts, but also run individually inside of workflows.

### DIFF
--- a/scripts/bdevperf.py
+++ b/scripts/bdevperf.py
@@ -1,0 +1,101 @@
+import json
+import logging as log
+from argparse import ArgumentParser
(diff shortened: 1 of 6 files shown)


karlowich (approved):
LGTM!

safl (approved):
This is excellent.

One comment, for future work, we should probably include running fio. It nothing else... then to demonstrate the issue of the software-overhead it entails...

PR #38, opened 2026-03-31
docs: expand background with GPU terminology and cross-references

karlowich:
Expands the background section with GPU/CUDA concepts needed to reason
  about device-initiated I/O, and improves consistency across the document.

  - Define peer-to-peer PCIe early in the PCIe section and use the term
    consistently throughout
  - Add a full CUDA section covering thread hierarchy, synchronization,
    streams, memory hierarchy, APIs, UVA, compute capability, and GPUDirect
  - Add a device-initiated I/O subsection under Massively Parallel Processing
    Units describing the PCIe mechanics of accelerator-driven storage access
  - Clean up the NVMe Controllers section opening
  - Add cross-references from introduction and architecture to the new
    background definitions

### DIFF
--- a/docs/src/architecture.md
+++ b/docs/src/architecture.md
@@ -68,7 +68,7 @@ commands.
   path independently once the queues are established.
 
 **Buffer placement** describes where data buffers and queue structures reside,
-and determines whether peer-to-peer (P2P) DMA is involved.
+and determines whether peer-to-peer (P2P) DMA (see {ref}`sec-pcie`) is involved.
 
 - *Host memory*: data buffers reside in host DRAM. The NVMe controller performs
   DMA to and from host memory. When an accelerator needs the data, a separate
--- a/docs/src/background.md
+++ b/docs/src/background.md
@@ -652,16 +652,13 @@ All NVMe specification documents are freely available upon ratification and
 distributed via the nvme express website. For details please to look in the
 specification documents, however, since these have now grown into many thousands
 of pages spread upon multiple documents, then see these subsections as a brief
-overview on theory of operations and NVMe initiatlization and command processing
+overview on theory of operations and NVMe initialization and command processing
 flows.
 
-Directly to the first concept an NVMe Controller, this is the entity which
-is accessible via a transport. Transports include, PCI, TCP, and RDMA. We are
-currently focused on the locally attached storage devices over PCI. Thus, thus
-initially we will cover how to initialize an NVMe controller over PCIe and touch
-upon PCIe infra. for it.
-
-On with it.
+An NVMe controller is the entity accessible via a transport. Transports include
+PCIe, TCP, and RDMA. We currently focus on locally attached storage devices over PCIe, with the
+intention to expand to other transports in the future. The following subsections
+cover controller initialization and command processing in that context.
 
 NVMe devices implement a command driven model in which the host constructs
 commands, places them in submission queues, and notifies the controller through
--- a/docs/src/background.md
+++ b/docs/src/background.md
@@ -462,11 +462,188 @@ accelerator integrated storage paths.
 (sec-massively-parallel-processing-units)=
 ## Massively Parallel Processing Units
 
-* SIMD, SPMD, and SIMT
+Modern accelerators achieve high throughput by executing thousands of lightweight
+threads concurrently across many parallel compute units. Understanding how these
+threads are organized, scheduled, and provided with data is essential for
+reasoning about device-initiated I/O paths and the memory placement
+constraints they impose.
 
-* Arithmetic vs logic
+(sec-cuda)=
+### CUDA
 
-* Memory bandwidth vs Compute Capability
+CUDA (Compute Unified Device Architecture) is NVIDIA's parallel computing
+platform and programming model {cite}`hwu2022pmpp`. It exposes GPU hardware through a C-like
+programming interface that allows software to describe parallel computations as
+functions called **kernels**, which are launched on the GPU and execute across
+many threads simultaneously.
+
+#### Thread Hierarchy and Streaming Multiprocessors
+
+CUDA organizes threads into a three-level hierarchy under the **Single
+Instruction, Multiple Threads (SIMT)** execution model. Unlike SIMD, where a
+single instruction operates on a fixed-width vector of data in a single thread,
+SIMT issues a single instruction to a group of threads, each maintaining
+independent register state and a program counter.
+
+**Threads** are the smallest unit of execution in the SIMT model. Each thread
+has private registers and a unique position within its containing block and
+grid, expressed as `threadIdx`, `blockIdx`, and `blockDim` built-in variables.
+Threads are the granularity at which work is assigned and at which register
+state is maintained.
+
+**Thread blocks** (also called Cooperative Thread Arrays, or CTAs) are groups of
+threads that execute on the same compute unit and may cooperate through shared
+memory and explicit barrier synchronization. A thread block can contain up to
+1024 threads. The number of thread blocks that reside concurrently on a compute
+unit is limited by register and shared memory consumption.
+
+**Warps** are the fundamental unit of SIMT scheduling. A warp consists of 32
+threads that the hardware issues and executes as a single unit. Threads within
+a warp execute in lockstep under normal conditions, but may diverge when
+control flow differs, in which case inactive branches are masked until threads
+reconverge. While one warp stalls on a memory operation, the scheduler issues
+instructions from another ready warp, keeping execution units occupied.
+
+**Grids** are collections of thread blocks that together constitute a single
+kernel launch. Thread blocks within a grid execute independently and may be
+scheduled on any SM in any order. This independence is what enables kernels to
+scale across GPUs with varying numbers of SMs without requiring coordination
+between blocks.
+
+The **Streaming Multiprocessor (SM)** is the primary compute unit of a CUDA GPU.
+Each SM contains a warp scheduler, a register file partitioned among its active
+warps, and a configurable shared memory and L1 cache partition.
+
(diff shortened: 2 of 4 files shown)


docs/src/background.md
```diff
@@ -446,11 +451,11 @@ domains at the level defined by the controller.
 Data movement in PCIe whether DMA or MMIO are sequences of Memory TLPs
 whose routing depends on which component performs the decode. Devices that
 generate address-routed TLPs rely on the RC unless a switch provides its own
-decode windows. Devices capable of ID-routed TLPs can perform peer routing
+decode windows. Devices capable of ID-routed TLPs can perform peer-to-peer routing
 independently. IOMMU translation, ATS caching, and ATC entries determine how
 devices obtain the addresses they use, but they do not alter the routing model.
-Together, RC mediated DMA, switch mediated peer DMA, and device routed peer DMA
-describe all data transfer paths allowable in PCI Express systems and form the
+Together, RC-mediated DMA, switch-mediated peer-to-peer DMA, and device-routed
```
safl: The commit does a bit more than described in the commit message, that is, it also changes: ``rc mediated``, ``switch mediated``, and ``device routed`` to use hyphens. It would be fine to extend the commit-message if it was also here, but since since "switch mediated" and "rc mediated" are also used for section-titles, then it would be preferable that it would be addressed in a separate commit, or not changed to use hyphens.

docs/src/background.md
```diff
@@ -462,11 +462,203 @@ accelerator integrated storage paths.
...
+constraints they impose.
+
+(sec-cuda)=
+### CUDA
+
+CUDA (Compute Unified Device Architecture) is NVIDIA's parallel computing
+platform and programming model. It exposes GPU hardware through a C-like
+programming interface that allows software to describe parallel computations as
+functions called **kernels**, which are launched on the GPU and execute across
+many threads simultaneously. CUDA implements a **Single Instruction, Multiple
```
safl: How about moving:

``CUDA implements a **Single Instruction, Multiple
Threads (SIMT)** execution model: unlike SIMD, where a single instruction
operates on a fixed-width vector of data in a single thread, SIMT issues a
single instruction to a group of threads, each maintaining independent register
state and a program counter. Threads within a group execute in lockstep under
normal conditions, but may diverge when control flow differs, in which case
inactive branches are masked until threads reconverge.``
(rephrased) into **Thread Hierarchy and Streaming Multiprocessors**?

E.g. the SIMT vs SIMD might be nice to have inside the **Threads** description, to support the description of what a thread is.

docs/src/background.md
```diff
@@ -462,11 +462,203 @@ accelerator integrated storage paths.
...
+`cudaHostRegisterIoMemory`, the two features central to device-initiated NVMe
+I/O, both require compute capability 3.5, introduced with the Kepler
+architecture in 2013. This requirement is satisfied by all modern CUDA-capable
+GPUs and is not a practical constraint. The binding limitation is instead
+platform-level peer-to-peer routing support, governed by PCIe topology and switch
+capabilities, as covered in {ref}`sec-pcie`.
+
+#### GPUDirect
+
+**GPUDirect** is a family of NVIDIA technologies that enable data paths between
```
safl: I am a bit wavy whether we want this here as it is already in the introduction... **NVIDIA GPUDirect Storage (GDS) {cite}`nvidia-gds` was the first widely deployed...** i currently lean towards dropping this subsection.

docs/src/background.md
```diff
@@ -660,6 +660,19 @@ three variants depend on UVA for coherent addressing, on peer-to-peer PCIe
 routing for the data transfer, and on the platform permitting direct BAR-to-BAR
 access between the participating endpoints.
 
+(sec-device-initiated-io)=
+### Device-Initiated I/O
+
+When an accelerator acts as the initiator of storage I/O, peer-to-peer PCIe
```
safl: I don't like this subsection.
First, ``...is involved in both directions of the transfer..`` in what transfer?
Second, `` Outbound, the accelerator writes commands and rings doorbells by issuing stores directly to the storage controller's BAR``. This reads as "the accelerator writes NVMe command submission queue entries in memory mapped in the storage controllers BAR"... however that is not accurate, that **can** happen when a feature like "submission-queues-in-CMB" is being used, but that is not the common case, and not what we are practicing. It muddies/mixed "mmio" and "DMA" roles. As I recall, then we go to great lengths later in defining these things more clearly. I would lean towards dropping this subsection.

docs/src/background.md
```diff
@@ -680,16 +680,13 @@ All NVMe specification documents are freely available upon ratification and
...
 flows.
 
-Directly to the first concept an NVMe Controller, this is the entity which
-is accessible via a transport. Transports include, PCI, TCP, and RDMA. We are
-currently focused on the locally attached storage devices over PCI. Thus, thus
-initially we will cover how to initialize an NVMe controller over PCIe and touch
-upon PCIe infra. for it.
-
-On with it.
+An NVMe controller is the entity accessible via a transport. Transports include
```
safl, line 658: Great.

docs/src/architecture.md
```diff
@@ -61,14 +61,15 @@ commands.
   queue entries, populates data transfer descriptors (PRPs or SGLs), and writes
   doorbell registers. This applies to both kernel managed and user space managed
   infrastructure.
-- *Device-initiated*: software executing on a PCIe-attached accelerator (such as
-  a GPU) constructs and submits NVMe commands from device-resident driver code.
+- *Device-initiated* (see {ref}`sec-device-initiated-io`): software executing on a
```
safl: I would drop this reference.

docs/src/architecture.md
```diff
@@ -61,14 +61,15 @@ commands.
...
+  PCIe-attached accelerator (such as a GPU) constructs and submits NVMe commands
+  from device-resident driver code.
   Host-side software remains responsible for device initialization, queue
   provisioning, and metadata resolution, but the accelerator drives the data
   path independently once the queues are established.
 
 **Buffer placement** describes where data buffers and queue structures reside,
-and determines whether peer-to-peer (P2P) DMA is involved.
+and determines whether peer-to-peer (P2P) DMA (see {ref}`sec-pcie`) is involved.
```
safl, line 71: great

docs/src/introduction.md
```diff
@@ -42,7 +42,7 @@ entirely.
 Independent of software overhead, accelerator workloads suffer from redundant
 data movement. When a GPU requires data from storage, the conventional path
 routes it through host DRAM: the NVMe controller writes to host memory, and a
-separate copy transfers the data to GPU memory. Peer-to-peer (P2P) DMA
+separate copy transfers the data to GPU memory. Peer-to-peer (P2P) DMA (see {ref}`sec-pcie`)
```
safl, line 45: great

docs/src/introduction.md
```diff
@@ -56,8 +56,8 @@ under development within the Linux kernel.
 
 ## Device-Initiated I/O
 
-A third, independent direction is device-initiated I/O, where the accelerator
-itself constructs and submits NVMe commands. This removes the host CPU from the
+A third, independent direction is device-initiated I/O (see {ref}`sec-device-initiated-io`),
```
safl: I would drop this reference.

safl (commented):
This is excellent. There is only one thing I disagree with and that is the **device-initiated-io** subsection added to the background section and referred to elsewhere. Otherwise, just some minor things inline.

docs/src/background.md
karlowich: That's an oversight. Didn't mean to add those hyphens.

docs/src/background.md
karlowich: I will drop it

docs/src/architecture.md
karlowich: Yes

docs/src/introduction.md
karlowich: Yes

docs/src/background.md
karlowich: The hyphen for device-routed is correct. The others have been removed

docs/src/background.md
karlowich: Yes, I'm also unsure. This establishes GPUDirect as bigger than GDS which I like. However, I don't know if it is really relevant as background. Perhaps it should go in related work instead. Or just be dropped.

docs/src/background.md
karlowich: Yes

docs/src/background.md
karlowich: I dropped it, we already have a section about this in related work

karlowich:
@safl Addressed all your concerns

docs/src/background.md
```diff
@@ -462,11 +462,188 @@ accelerator integrated storage paths.
...
+Modern accelerators achieve high throughput by executing thousands of lightweight
+threads concurrently across many parallel compute units. Understanding how these
+threads are organized, scheduled, and provided with data is essential for
+reasoning about device-initiated I/O paths and the memory placement
+constraints they impose.
+
+(sec-cuda)=
+### CUDA
+
+CUDA (Compute Unified Device Architecture) is NVIDIA's parallel computing
```
safl, line 474: When reading this, it might be nice with some pointers on where to read more... perhaps drop a citation of the classic **Programming Massively Parallel Processors: A Hands-on Approach** here.

safl (approved):
I had one comment that I did not get into previous review... adjust if you think it makes sense but these changes are great, fixing slob and great additions in the missing sections, excellent.

docs/src/background.md
karlowich: Done

safl (approved):
Excellent.

PR #39, opened 2026-04-01
bench: transition from libnvm/BAM to xNVMe uPCIe backends

karlowich:
Completes the transition from the libnvm/BAM stack to xNVMe with uPCIe
  backends, and cleans up the nvstack configuration.

  Benchmark changes:
  - Switch dataset benchmarks (filesize8gib, tiktokish, imagenetish) from
    the aisio backend to aisio-cpu, replacing libnvm device references
    with PCI addresses and uio_pci_generic driver binding
  - Replace sil seqread/randread synthetic benchmarks with xnvmeperf
    randread using the upcie and upcie-cuda backends
  - Simplify sil.py to single-backend execution

  Configuration changes:
  - Switch xNVMe from a fork to upstream xnvme/xnvme on the next branch
  - Drop bam and upcie as standalone build components
  - Replace libnvm_dev with pci_addr in dataset config
  - Bump sil to v0.2.0

  nvstack fixes:
  - Move hardcoded CUDA pin URL into config
  - Use nokm source tree when building nvidia-fs
  - Drop unused doca config entry


naddinadja (approved):
great, Karl! :)

scripts/sil.py
```diff
@@ -48,17 +48,12 @@ def get_opts(args, cijoe, backend):
...
-        dataset = f"--data-dir {args.dataset}"
-
-    for backend in args.backends:
-        opts = get_opts(args, cijoe, backend)
-        cmd = f"{prefix} {args.bin} {args.device} {dataset} --batches {args.batches} --batch-size {args.batch_size} --backend {backend} {opts}"
-        for _ in range(args.repetitions):
-            err, _ = cijoe.run(cmd)
-        if err:
-            return 1
+    opts = get_opts(args, cijoe, args.backend)
```
fmma, line 48: get_opts() has no case for "aisio-cpu", so it falls through and returns None. This appends the literal string "None" to the sil command, which SIL rejects as an unexpected argument. Add an aisio-cpu case that returns the empty out string.

tasks/bench_aisio.yaml
```diff
@@ -21,45 +20,45 @@ steps:
...
+- name: upcie
+  uses: xnvmeperf
   with:
-    backends: ["aisio"]
-    dataset: ""
-    device: "{{ config.filesystems.dset.libnvm_dev }}"
-    batch_size: 16777216
-    random: false
+    backend: "upcie"
+    device: "{{ config.filesystems.dset.pci_addr }}"
```
fmma: xnvmeperf.py defines --devices (plural, nargs="+") but this passes device (singular). The value is silently ignored, args.devices stays [], and main() errors out. Either rename to devices here (and in the upcie-cuda step), or add a --device parameter to xnvmeperf.py.

tasks/bench_aisio.yaml
```diff
@@ -21,45 +20,45 @@ steps:
...
+- name: upcie-cuda
+  uses: xnvmeperf
   with:
-    backends: ["aisio"]
-    dataset: ""
-    device: "{{ config.filesystems.dset.libnvm_dev }}"
-    batch_size: 16777216
-    random: true
+    backend: "upcie-cuda"
+    device: "{{ config.filesystems.dset.pci_addr }}"
```
fmma: same

fmma (commented):
I ran bench_aisio after setup_dataset and hit intermittent failures in the tiktokish step. For some reason, a handful of files ended up with 2 extents after setup dataset. SIL does not seem to support files with fragmentation yet. Running xfs_fsr on the mounted filesystem before rebinding to uio_pci_generic fixed it. Maybe add a defrag step at the end of setup_dataset?

scripts/sil.py
karlowich: Thanks

tasks/bench_aisio.yaml
karlowich: Thanks

tasks/bench_aisio.yaml
karlowich: Thanks

karlowich:
@fmma I have addressed your comments now

safl (commented):
Since this marks a nice reset, how about adding a small fix to ``README.md`` such that it starts by recommending the installation of Ubuntu as described in ``tasks/setup_ubuntu.yaml`` making the ``README.md`` a true end-to-end guide to start from scratch?

Also, since this is the "end-of-poc" this could marked in the ``README.md`` as well by removing poc from: ``This repository provides documentation, tools, and scripts for setting up and exploring the Accelerator-integrated Storage I/O Proof-of-Concept.`` something like:
```
This repository provides documentation, tools, and scripts for setting up and experimenting with Accelerator-integrated Storage I/O.
```
---
**EDIT-NOTE:** I kept saying ``README.md`` completely forgetting that this is still restructured text... might be worth a makeover as well :)

karlowich:
@safl I think that's a good point, will do

README.md
```diff
@@ -0,0 +1,151 @@
...
+exploring Accelerator-integrated Storage I/O. The guide below walks through the
+full setup from a fresh Ubuntu installation to running `bench_aisio.yaml`.
+
+## Prerequisites
+
+The following tools are used throughout this guide. Install them with `pipx`:
+
+```
+pipx install cijoe
+pipx install devbind
```
safl: ``devbind`` and ``hugepages`` should be installed on the target system, not on the "initiator" e.g. not on the system that has ``cijoe``, thus, remove it here and at it somewhere inside of ``setup_aisio.yaml``.

safl (changes requested):
This is looking really good. One small thing inline.

README.md
karlowich: Right of course

fmma (approved):
Fixes look correct.

safl (approved):
This is awesome.

PR #42, opened 2026-04-07
Tools experiment: automation and docs

naddinadja:
A PR for the automation of running and visualising the tools experiment, and also the docs describing the experiment and results.

bench_tools-results


docs/src/results.md
```diff
@@ -208,3 +208,97 @@ The results scale linearly with the amount of devices used.
...
+### NVMe Driver Comparison
+
+Since all configurations in this section use **xnvmeperf**, backends are referred
+to by name alone for brevity.
+
+With the benchmark tool and xNVMe abstraction layer held constant, the **upcie**
+backend reaches approximately 2.5 times the IOPS of the **spdk** backend across
+both thread counts. As noted in {ref}`sec-experiments-tool-comparison`, this
+comparison reflects differences in NVMe driver implementation rather than
+abstraction depth. The **spdk** backend relies on SPDK, which routes completions
```
karlowich: This is not exactly true. It is definitely also due to the SPDK NVMe driver suffering from heavy abstraction which adds extra cost to every call. Maybe you can remember, I made a single indirection layer in uPCIe which added %5 overhead costing us millions of IOPS.

docs/src/results.md
```diff
@@ -208,3 +208,97 @@ The results scale linearly with the amount of devices used.
...
+record a timestamp, and on every completion it calls ``spdk_get_ticks()`` again
+to compute the elapsed time, then updates per-queue minimum, maximum, and total
+TSC accumulators, and optionally records the sample in a histogram. In addition,
+``spdk_get_ticks()`` is called on every iteration of the polling loop in
+``work_fn``. ``spdk_get_ticks()`` maps to ``rdtsc``, a non-trivial instruction
+that cannot be freely reordered; at tens of millions of IOPS the cumulative cost
+of issuing it twice per I/O plus once per poll iteration becomes measurable.
+
+Removing these calls from **perf** yields approximately 20% higher IOPS at one
+thread, narrowing the gap to **xnvmeperf** to approximately 16%. The improvement
```
karlowich, line 270: Are you sure we removed all the `get_ticks()`? I recall that we got better numbers, but maybe I'm mistaken

docs/src/experiments.md
```diff
@@ -244,35 +241,104 @@ The three tools under comparison are:
...
-devices bound to user space drivers. The comparison demonstrates differences in
-achievable IOPS attributable to the benchmark tool overhead and the underlying
-user space NVMe driver implementation.
+Both configurations use the same benchmark tool, the same xNVMe abstraction
+layer, and the same uPCIe NVMe driver. The only structural difference is buffer
+placement: the upcie backend uses host memory, while the upcie-cuda backend
+allocates data buffers in GPU device memory and transfers data via peer-to-peer
+DMA over the PCIe fabric, bypassing host DRAM entirely. Performance differences
+between these two configurations therefore reflect the effect of P2P buffer
+placement on the data path, independent of the NVMe driver or benchmark tool.
```
karlowich, line 322: These sections are like extensions of the "I/O Paths", perhaps they can be combined?

karlowich (changes requested):
Code looks good. Some comments to the docs.

safl (approved):
lgtm

docs/src/results.md
naddinadja: I removed more than when we tested it, but no we didn't get different numbers. Went from 10 to 12 millions

docs/src/experiments.md
naddinadja: I don't think I understand. Here we first have the 5 different I/O paths described, and directly afterwards how we can (and will) compare them. Do you want to mix that?

safl: 2 cents, this looks good to me since these are descriptions of how parameter sets in the experiment are labelled and compared.

docs/src/results.md
naddinadja: I've clarified that it's the xNVMe abstraction that does not differ now :)

docs/src/experiments.md
karlowich: All good then :)

karlowich (approved):
Great!

PR #43, opened 2026-04-07
docs: add Experimental Framework section and expand environment descriptions

karlowich:
- Adds a new Experimental Framework section covering system setup,
    software installation order, and benchmark workflows (synthetic and
    file-based), extracted from the Environments section where it did not
    belong
- Expands Environments with full hardware tables: descriptive system
    names, class descriptions, motherboard, CPU, memory, storage, and PCIe
    generation constraints for all four setups
- Centralises all NVIDIA software versions in nvstack.toml; install
    scripts (setup_nvstack.yaml, setup_nvstack_from_source.yaml) now
    reference config values instead of hardcoding versions, and the
    white-paper sources the same values dynamically via conf.py
- Updates SPDK to the official repository at tag 26.01

### DIFF
--- a/configs/aisio.toml
+++ b/configs/aisio.toml
@@ -18,8 +18,8 @@ tag = "v0.2.0"
 path = "/root/git/sil"
 
 [spdk.repository]
-remote = "https://github.com/ankit-sam/spdk.git"
-branch = "xnvme_bam"
+remote = "https://github.com/spdk/spdk"
+tag = "26.01"
 path = "/root/git/spdk"
 
 [fio.repository]
--- a/docs/src/experimental_framework.md
+++ b/docs/src/experimental_framework.md
@@ -1,0 +1,236 @@
+(sec-experimental-framework)=
+# Experimental Framework
+
+## System Setup
+
+All environment provisioning is automated using
+[CIJOE](https://github.com/refenv/cijoe), a tool for systems development
+and testing. CIJOE runs on a local initiator machine and connects to a
+remote target over SSH. Shell commands and Python scripts are collected
+into YAML workflow definitions that document the execution sequence and
+combined purpose. Input values are kept separate from scripts in TOML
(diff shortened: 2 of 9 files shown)


docs/src/experimental_framework.md
```diff
@@ -0,0 +1,236 @@
...
+configuration files, so the same workflow can be replicated across
+different environments without modifying the scripts themselves. After
+execution, CIJOE generates a report covering command output, script
+documentation, collected artifacts, and a workflow summary.
+
+Step-by-step instructions for running the CIJOE setup tasks are provided
+in the ``README.md`` of the AiSIO repository. The table below lists all
+software components installed on each system, in installation order.
+
+| Category | Component                  | Version              | Install       | Description                                                                                                                         |
```
safl, line 21: Really awesome how the version info is extracted directly from the config. So good.

safl (approved):
lgtm(very)!

configs/aisio.toml
```diff
@@ -18,8 +18,8 @@ tag = "v0.2.0"
 path = "/root/git/sil"
 
 [spdk.repository]
-remote = "https://github.com/ankit-sam/spdk.git"
-branch = "xnvme_bam"
+remote = "https://github.com/spdk/spdk"
+tag = "26.01"
```
yonggilsong, line 22: I hit a setup failure in the SPDK checkout step with the current config. repository_prep ends up running git checkout 26.01, but that ref was not found in the SPDK repo in my test environment, so setup stopped there. After changing the configured tag to v26.01, setup_aisio.yaml completed successfully. It looks like the intended SPDK ref should be v26.01 rather than 26.01.

karlowich: Yes you're absolutely right. That's my mistake

karlowich: Merged into main

PR #46, opened 2026-04-08
chore: rename xallib to xal after repo move to xnvme/xal

safl:


naddinadja (approved):
perfect!

karlowich (approved):
Great let's do the same for SIL

PR #53, opened 2026-04-15
Artifacts - data for plots

naddinadja:
We have four main plots:
1. `barplot-sat`: A bar plot of saturating PCIe with upcie-cuda backend
2. `barplot-tool`: A bar plot with the comparison of the different tools and drivers
3. `lineplot-spdk`: A line plot with physical cores on the x-axis and IOPS on the y-axis, showing how bdevperf+spdk scales when physical cores increases
4. `lineplot-upcie`: Same as above, just with xnvmeperf+upcie and xnvmeperf+upcie_cuda

This PR adds the cijoe scripts that reads the results files from the remote server and converts the data into a yaml format useable in matplotlib-plotting scripts. The PR also has the actual results in the right format.

PR #52, opened 2026-04-09
feat(bench): add PCIe bandwidth benchmark with GPU monitoring

karlowich:
- Adds `DcgmHelper`, a new script that starts `dcgmi dmon` in the
    background during a benchmark run and returns per-field statistics
    (mean, p95, and raw samples) for post-processing
  - Integrates `DcgmHelper` into `BenchHelper` for `upcie-cuda` runs,
    storing dcgm field data alongside IOPS and bandwidth in the result JSON
  - Adds `tasks/bench_pcie.yaml`, a dedicated task that runs `xnvmeperf`
    with the `upcie-cuda` backend across multiple I/O sizes to characterise
    PCIe bandwidth during P2P transfers to GPU memory
  - Fixes a pre-existing missing return tuple (`-1` → `-1, None`) in the
    unknown-tool branch of `BenchHelper.run_benchmark`

PR #54, opened 2026-04-21
feat(docs): add sections on the pcie bw saturation experiment

naddinadja:
Please note that this section describes the experiment developed in #52 and use the data and plot from #53

### DIFF
--- a/docs/src/experiments.md
+++ b/docs/src/experiments.md
@@ -266,3 +266,64 @@ The independent variables are:
 ### Results
 
 Results are presented in {ref}`sec-results-tool-comparison`.
+
+(sec-experiments-pcie-bandwidth)=
+## PCIe Bandwidth Characterisation
+
+With the **upcie-cuda** driver, NVMe command data payloads are transferred directly
+to GPU device memory via P2P DMA, and therefore we want to inspect how much of the
+available PCIe bandwidth the data path consumes relative to the link capacity.
+Application-level benchmarks such as xnvmeperf report payload bytes, but do not
+account for the NVMe and PCIe protocol traffic that accompanies each transfer. As
+such, we also want to evaluate whether the payload bandwidth reported by the
+benchmark tool agrees with what the PCIe hardware counters observe.
+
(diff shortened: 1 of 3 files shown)


docs/src/experiments.md
```diff
@@ -266,3 +266,64 @@ The independent variables are:
 ### Results
 
 Results are presented in {ref}`sec-results-tool-comparison`.
+
+(sec-experiments-pcie-bandwidth)=
+## PCIe Bandwidth Characterisation
+
+With the **upcie-cuda** driver, NVMe command data payloads are transferred directly
+to GPU device memory via P2P DMA, and therefore we want to inspect how much of the
+available PCIe bandwidth the data path consumes relative to the link capacity.
```
karlowich, line 275: I think goal is to see how little CPU is needed to saturate the GPUs PCIe link. The data vs overhead is a by-product of the original intention.

We only have an influence on the first part. The only thing we can do about the PCIe overhead is try to explain it.

Is this making sense?

karlowich (commented):
Just one comment about the framing. This bleeds into all 3 files, but didn't want to comment the same everywhere.

I'm unsure about whether we want to change the framing. As such, I won't approve or request changes.

docs/src/experiments.md
safl: Right, we want to show once we switch to a bandwidth focus, instead of IOPS, then CPU utilization goes to zero. Eventually this will translate into observations like payloads at MTDS will utilize close to zero CPU. E.g. with payloads of e.g. 256K we need 1/512 the amount of IO processing... so we go from needingto process 62M IOPS down to 121K IOPS. Having said that, then the surprising thing that one probably want to have an affirmation of is why aren't we just hammering out 64GB/sec... e.g. we want to clearly demonstrate that the lost BW is due to transport-overhead, not software-overhead. So for that reason, I am good with grabbing these sections as is, and then once we have this PR in and https://github.com/xnvme/aisio/pull/53 then we can re-iterate on the paper holistically.

safl (approved):
lgtm.

PR #56, opened 2026-04-22
Add device-initiated I/O benchmarking with xnvmeperf-cuda

karlowich:
Adds end-to-end support for benchmarking device-initiated I/O using
  xnvmeperf's `cuda-run` subcommand with the `upcie-cuda` backend.

  ## Changes

  **Bench tooling** (`feat(bench)`)
  Extend `bench_runall`, `bench_helper`, and `bench_combine` to support
  xnvmeperf-cuda as a tool and `--nqueues` as a sweep dimension. GPU runs
  use a streamlined result filename; `bench_combine` handles both CPU and
  GPU formats so all results flow through the same visualisation pipeline.

  **Queue depth scaling experiment** (`feat: add bench_cuda_qdepth ...`)
  Sweeps queue depth over {1–512} with nqueues per device as the secondary
  variable {1–16}, fixing I/O size at 512 B across 16 NVMe devices. Includes
  cijoe task, plot script, Jinja2 template, and results artifact.

  **I/O size scaling experiment** (`feat: add bench_cuda_iosize ...`)
  Sweeps I/O size with queue depth as the secondary variable, fixing nqueues=1
  across 16 devices. Primary metric is bandwidth rather than IOPS. Includes
  cijoe task, plot script, Jinja2 template, and results artifact.

  **Plot tooling** (`feat(plots)`)
  - Add bandwidth lineplot support: rooflines now detect `value_nbytes` vs
    `value_iops` and scale the y-axis to GiB/s or millions of IOPS accordingly
  - Apply plasma colormap to cuda lineplots; general lineplot visual improvements

  **Documentation** (`docs`)
  - Expand implementation section to describe both I/O modes (CPU-initiated P2P
    and device-initiated), document `xnvme_cuda_queue_create()`, and detail the
    four-stage CUDA kernel submission flow in `xnvme_cuda_cmd_io()`
  - Remove the "Device-Resident NVMe Driver" future work section — the blocking
    re-implementation is complete
  - Update "Evaluating Device-Initiated Paths" to reflect initial benchmarking
    is in place; identify FIL integration as the next step
  - Add results analysis and reorganize results sections

  **Artifacts**
  Migrate existing binary artifacts to Git LFS; add new artifacts for both
  cuda experiments.

  ## Graphs
  
  ### Scaling IO size
lineplot-cuda-iosize

  ### Scaling qdepth
lineplot-cuda-qdepth


scripts/bench_combine.py
```diff
@@ -34,20 +34,27 @@ def main(args, cijoe: Cijoe):
...
+        if m:
+            ht, st, sm, tu = map(int, m.groups())
+            label = (
+                f"{'U' if ht else 'Not u'}sing thread siblings; "
+                f"SMT {'on' if sm else 'off'}; "
+                f"stress {'on' if st else 'off'}; "
+                f"turbo {'on' if tu else 'off'}"
+            )
+        else:
+            label = f"xnvmeperf-cuda; nqueues {int(cuda_m.group('nq'))}"
```
naddinadja: I would prefer this to be an `if m: (...) elif cuda_m: (...) else: log.error (......)`. It would make it more clear that the `else` here refers to the cuda situation.

tasks/bench_pcie.yaml
```diff
@@ -62,6 +62,17 @@ steps:
...
+  uses: bench_runall
+  with:
+    depths: *qd
+    sizes: *io
+    numdevs_specific: *devs
+    nqueues: [1, 2, 4]
+    results_dir: *results-dir
+    tool: xnvmeperf-cuda
+    backend: upcie-cuda
+
```
naddinadja: Why add this here? Does it add an angle to the experiment that we are missing when testing with P2P?

tasks/bench_tools.yaml
```diff
@@ -124,6 +124,17 @@ steps:
     tool: xnvmeperf
     backend: upcie-cuda
 
+- name: xnvmeperf_cuda
+  uses: bench_runall
+  with:
+    depths: *qd
+    sizes: *io
+    numdevs_specific: *devs
+    nqueues: [1, 2, 4]
```
naddinadja: This will probably break the visualisations, as the graphs we use for this experiment are barplots categorised by the run's tool+backend combination. I have not tested it, but I assume it will try to place all 3 bars on top of each other if there is not added any logic for separating them.

naddinadja (changes requested):
Comments and questions inline :)

tasks/bench_tools.yaml
naddinadja: Oh wait no, the x-axis here is currently "Number of CPU threads". We can change it to "Number of threads" instead and specify that "threads" refer to CPU threads in CPU initiated runs and GPU threads in GPU initiated runs. However - we have 1,2,3 and 4 CPU threads in the other experiments, do we want the same here?

naddinadja: I have resolved the part of it breaking the visualisations, but we still need to run it with 1,2,3,4 nqueues - I didn't want to touch your commits, so you get to do this :)

tasks/bench_pcie.yaml
naddinadja: As we talked about, no it does not, so we remove this from here, and define another experiment in another PR exercising the GPU-initiated path.

tasks/bench_tools.yaml
karlowich: I have dropped this functionality

tasks/bench_pcie.yaml
karlowich: I have dropped this

scripts/bench_combine.py
karlowich: I will fix this

karlowich:
This has been revamped. Please take a look at the new description.

docs/src/experiments.md
```diff
@@ -327,3 +327,58 @@ arithmetic means.
...
 {ref}`sec-experimental-framework`.
+
+(sec-experiments-cuda-qdepth)=
+## Device-initiated I/O: Queue Depth Scaling
+
+The previous experiments fix queue depth at 128 and vary I/O size or I/O path.
+This experiment instead holds I/O size fixed and sweeps queue depth across a wide
+range, with the number of queues per device as the secondary variable. The goal is
+to characterise how IOPS scales with queue depth under device-initiated I/O and to
+identify the depth at which each queue-count configuration saturates.
```
naddinadja: This introduction kinda makes it sound like the main difference between this exp and the previous ones is the way we sweep across the parameter set - but isn't the "main main" difference that this is done on a device-initiated path?

docs/src/results.md
```diff
@@ -342,3 +342,21 @@ exceeds the payload bandwidth reported by xnvmeperf by a consistent factor of
...
+
+(sec-results-cuda-qdepth)=
+## Device-initiated I/O: Queue Depth Scaling
+
+This section presents the results of the experiment described in
+{ref}`sec-experiments-cuda-qdepth`.
+
+Results are presented as IOPS vs. queue depth, with one line per number of queues
+(``nqueues`` ∈ { 1, 2, 4, 8, 16 }). All configurations use **xnvmeperf** with the
+``cuda-run`` subcommand and the **upcie-cuda** backend, 16 NVMe devices, and 512-byte I/O.
```
naddinadja: "This section presents results ..." and then "Results are presented ...." sound kinda borky
Also, do you want to do a little bit of explanation (and reflection) on what the graph shows? I know in a regular academic paper, you would keep results and discussion separate, but we have merged it a bit in our results section. At least, I feel like you should be able to understand the main results without the graph too.

naddinadja: Love the "∈"

docs/src/experiments.md
```diff
@@ -267,6 +267,65 @@ The independent variables are:
 
 Results are presented in {ref}`sec-results-tool-comparison`.
 
+(sec-experiments-cuda-iosize)=
+## Device-initiated I/O: I/O Size Scaling
+
+The previous queue depth experiment fixes I/O size at 512 bytes and sweeps queue
+depth to characterise IOPS scaling. This experiment inverts the approach: it holds
+queue depth as the secondary variable and sweeps I/O size across a wide range, from
```
naddinadja: Are you referring to the previous Device-Initiated exp? Because then this should probably be placed *after* that section hehe

docs/src/results.md
```diff
@@ -299,6 +299,27 @@ routing the data path through P2P DMA to GPU device memory does not measurably
...
+## Device-initiated I/O: I/O Size Scaling
+
+This section presents the results of the experiment described in
+{ref}`sec-experiments-cuda-iosize`.
+
+Results are presented as PCIe RX bandwidth vs. I/O size, with one line per queue depth
+(``qdepth`` ∈ { 1, 2, 4, 8, 16, 32, 64, 128 }). All configurations use **xnvmeperf**
+with the ``cuda-run`` subcommand and the **upcie-cuda** backend, ``nqueues=1``, and 16
+NVMe devices. Total CUDA thread count equals queue depth × 16. The dashed reference
+line marks the peak P2P bandwidth from ``p2pBandwidthLatencyTest``.
```
naddinadja: Same comment as the previous results section

docs/src/results.md
naddinadja: I see that the description comes in a later commit - idk if it's necessary, but we COULD move it so that they're in the same commit

docs/src/results.md
```diff
@@ -364,6 +343,59 @@ approximately 1.28. The consistency of this ratio across I/O sizes and IOPS leve
...
+```
+
+The results confirm the central hypothesis of the experiment: the queue depth required to
+saturate the PCIe link decreases monotonically as I/O size grows. At 512-byte I/O, even
+the maximum configuration of 2048 CUDA threads (``qdepth=128``, 16 devices) delivers only
+21.6 GB/s — the workload is constrained by per-device IOPS limits, not by available PCIe
+bandwidth. As I/O size increases, each command carries more payload, and the saturation
+threshold drops accordingly.
+
+The curve families converge at a practical ceiling of approximately 44–45 GB/s, which
```
naddinadja: What is a curve family?

naddinadja (changes requested):
Some comments inline :)

docs/src/results.md
karlowich: This is fixed now

docs/src/results.md
karlowich: This is fixed now

docs/src/experiments.md
karlowich: I have cleaned this up

docs/src/experiments.md
karlowich: Fixed

docs/src/results.md
karlowich: I have removed this term

naddinadja (approved):
looks good to me :)

PR #57, opened 2026-04-24
feat(artifacts): migrate binary artifacts to LFS

karlowich:
The artifact tarballs were previously committed as raw binary blobs directly in the object store (~11 MB total), despite .gitattributes declaring artifacts/** as LFS-tracked. This replaces the blobs with proper LFS pointers, so future clones and checkouts only download artifact data on demand.


karlowich:
Closing this as it is included here: #56

PR #58, opened 2026-04-27
Small fixes to provisioning of AiSIO

naddinadja:
I just tested the whole provisioning of AiSIO and found some small things that needed to be fixed :)


karlowich (approved):
LGTM

PR #61, opened 2026-04-28
Refactorization of Experiments section

naddinadja:
This PR refactors the docs such that the experiments and results are "merged". Instead of having all experiments described and then all results afterwards, we present the experiment results right after its description. This improves the readability of the report as a whole, but it also helps build the story of the paper, where the results of early experiments shape the definition of the following experiments.

With this structure, a bit of introduction and wrap-up was needed in each of the experiment sections, so later commits address this. Furthermore, the introduction lacked an emphasis on what we were trying to achieve in the paper, which was necessary to help establish the necessity of the experiments.


karlowich (approved):
Awesome!

PR #64, opened 2026-05-05
feat(bench/compare): add fio xnvme compare workflow

yonggilsong:
## Summary

Add an AiSIO compare benchmark workflow for `fio + xNVMe` so we can compare
the `spdk` and `upcie` backends under the same benchmark matrix.

This series adds:
- `tasks/bench_fio_compare.yaml`
- `fio_xnvme` support in the benchmark path
- raw result collection into a single compare JSON
- compare HTML visualization
- validation and smoke coverage
- safer CPU control handling when `cpupower`, turbo sysfs, or CPU frequency
  sysfs entries are unavailable
- quoted `PCI_BLACKLIST` handling for multi-BDF device configs

## Usage

Setup order:

    cijoe --monitor -c configs/transport.toml tasks/setup_ubuntu.yaml
    cijoe --monitor -c configs/transport.toml -c configs/udmabuf_import.toml tasks/setup_udmabuf_import.yaml
    cijoe --monitor -c configs/transport.toml -c configs/nvstack.toml tasks/setup_nvstack.yaml
    cijoe --monitor -c configs/transport.toml -c configs/aisio.toml tasks/setup_aisio.yaml

Run the compare benchmark:

    cijoe --monitor \
      -c configs/transport.toml \
      -c configs/bench_fio_compare.toml \
      -c configs/devices_*.toml \
      tasks/bench_fio_compare.yaml

Results are written under `cijoe-output/artifacts/fio-compare/<bdf>/`.

Current defaults are `runtime=5s` and `fio_size=16GiB`.

## TODO

- Add a standalone `fio_xnvme` script with `add_args()` / `main()`
- Reduce duplication in `bench_fio_compare.yaml`
- Revisit platform-specific backend/device binding behavior
- Improve CPU frequency support detection on systems without standard sysfs exposure

### DIFF
--- a/scripts/bench_helper.py
+++ b/scripts/bench_helper.py
@@ -32,6 +32,7 @@ import logging as log
 
 from bdevperf import bdevperf_cmd, create_config as bdevperf_config
 from dcgm_helper import DcgmHelper
+from fio_xnvme import fio_xnvme_cmd
 from spdk_nvme_perf import spdk_nvme_perf_cmd
 from xnvmeperf import xnvmeperf_cmd, xnvmeperf_cuda_cmd
 
@@ -45,6 +46,7 @@ class BenchHelper():
             cfm: CpuFrequencyHelper,
             tool: str = "bdevperf",
             backend: str = "spdk",
+            fio_size: str = "16GiB",
     ):
         self.initialised = False
 
@@ -55,7 +57,7 @@ class BenchHelper():
         self.stress = False
         self.backend = backend
         self.tool = tool
-
+        self.fio_size = fio_size
         self.dcgm = DcgmHelper(cijoe) if backend == "upcie-cuda" else None
 
         self.use_thrsib = False
@@ -79,8 +81,11 @@ class BenchHelper():
                 self.bin = Path(spdk_path)  / "build" / "bin" / "spdk_nvme_perf"
         elif tool in ["xnvmeperf", "xnvmeperf-cuda"]:
             self.bin = "xnvmeperf"
+        elif tool == "fio_xnvme":
+            self.bin = "fio"
         else:
             log.error(f"Failed: Unknown tool({tool})")
+            return
 
         self.initialised = True
 
@@ -98,15 +103,15 @@ class BenchHelper():
             return err
 
         self.use_thrsib = use_thrsib
-
         return 0
 
-    def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int, cpu_freq: float, suffix: str = "", nqueues: int = 1):
+    def run_benchmark(self, rw: str, depth: int, size: int, ndevs: int, ncpus: int, time: int, cpu_freq: float, suffix: str = "", nqueues: int = 1):
         if not self.initialised:
             log.error("Failed: benchmarker not initialised correctly")
             return 1, None
 
         is_cuda = self.tool == "xnvmeperf-cuda"
+        is_fio = self.tool == "fio_xnvme"
 
         if is_cuda:
             filename = (
@@ -114,10 +119,18 @@ class BenchHelper():
                 f"be_{self.backend}-tool_{self.tool}"
                 f"{suffix}.out"
             )
+        elif is_fio:
+            filename = (
+                f"d{ndevs}-c{ncpus}-o{size}-f_{self.fio_size}-"
+                f"rw_{rw}-q{depth}-be_{self.backend}-tool_{self.tool}-"
+                f"thrsib{1 if self.use_thrsib else 0}-"
+                f"freq_{cpu_freq}-"
+                f"stress{1 if self.stress else 0}"
+                f"{suffix}.out"
+            )
         else:
             filename = (
-                f"d{ndevs}-c{ncpus}-o{size}-q{depth}-"
-                f"be_{self.backend}-tool_{self.tool}-"
+                f"d{ndevs}-c{ncpus}-o{size}-q{depth}-be_{self.backend}-tool_{self.tool}-"
                 f"thrsib{1 if self.use_thrsib else 0}-"
                 f"freq_{cpu_freq}-"
                 f"stress{1 if self.stress else 0}"
@@ -131,15 +144,18 @@ class BenchHelper():
                 return 0, result
 
         bench_args = {
-            "iopattern": "randread",
+            "iopattern": rw,
             "qdepth": depth,
             "iosize": size,
             "runtime": time,
             "devices": [d["pci_addr"] for d in self.devices[0:ndevs]],
         }
 
-        if not is_cuda:
+        if is_cuda:
+            selected_cpus = []
+        else:
             bench_args["cpumask"] = self.cpu_masks[ncpus]
+            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
 
         command = f"/usr/bin/time "
 
@@ -150,10 +166,8 @@ class BenchHelper():
 
             bench_args["config_path"] = self.remote_config
             command += bdevperf_cmd(self.bin, bench_args)
-
         elif self.tool == "spdk_nvme_perf":
             command += spdk_nvme_perf_cmd(self.bin, bench_args)
-
         elif self.tool == "xnvmeperf":
             bench_args["backend"] = self.backend
             command += xnvmeperf_cmd(self.bin, bench_args)
@@ -163,14 +177,20 @@ class BenchHelper():
             bench_args["nqueues"] = nqueues
             command += xnvmeperf_cuda_cmd(self.bin, bench_args)
 
+        elif is_fio:
+            if ndevs != 1:
+                log.error("Failed: fio_xnvme currently supports exactly 1 device per benchmark point")
+                return 1, None
+
+            bench_args["backend"] = self.backend
+            bench_args["rw"] = rw
+            bench_args["fio_size"] = self.fio_size
+            cpu_list = ",".join(str(cpu) for cpu in selected_cpus)
+            command = f"taskset -c {cpu_list} {command}"
+            command += fio_xnvme_cmd(self.bin, bench_args)
         else:
             log.error(f"Unknown tool: {self.tool}")
             return -1, None
-
-        if is_cuda:
-            selected_cpus = []
-        else:
-            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
 
         if self.stress and (stressed_cpus := [str(x) for x in range(len(self.cpu_pairs)) if x not in selected_cpus]):
             command = "\n".join([
@@ -234,19 +254,23 @@ class BenchHelper():
         cpu_freqs = [[cpu_freqs[idx], self.cpu_pairs[idx]] for idx in selected_cpus]
 
         result = {
+            "rw": rw,
             "qdepth": depth,
             "iosize": size,
+            "fio_size": self.fio_size,
             "ndevs": ndevs,
             "ncpus": ncpus,
             "nqueues": nqueues,
+            "device_bdf": bench_args["devices"][0] if bench_args["devices"] else None,
             "cpu_usage": cpu_usage,
             "cpu_freqs": cpu_freqs,
             "fixed_freq": self.cfm.fixed_freq,
             "cpu_governor": self.cfm.governor,
-            "thr_sib": 1 if self.use_thrsib else 0,
+            "cpu_control_supported": 1 if self.cfm.cpu_control_supported else 0,
+            "thr_sib": self.use_thrsib,
             "smt": 1 if "SMT1" in suffix else 0,
             "turbo": 1 if "turbo1" in suffix else 0,
-            "stress": 1 if self.stress else 0,
+            "stress": self.stress,
             "tool": self.tool,
             "backend": self.backend,
             "iops": bench_result["total"]["iops"],
@@ -271,7 +295,6 @@ class BenchHelper():
             `(err, cpu_masks)` where a non-zero value for `err` describes that an error
             occured either while running `lscpu` or while parsing the output.
         """
-
         err, state = self.cijoe.run("lscpu -e")
         if err:
             log.error(f"Failed: lscpu -e")
@@ -284,7 +307,6 @@ class BenchHelper():
             return 1,
 
         cpu_pairs = [[int(v) for v in match.groupdict().values()] for match in matches]
-
         if use_thrsib:
             pairs = sorted(cpu_pairs, key=lambda p: p[1])
         else:
@@ -305,7 +327,6 @@ class BenchHelper():
 
         self.cpu_masks = cpu_masks
         self.cpu_pairs = cpu_pairs
-
         return 0
 
     def _parse_bench_results(self, table: str) -> Tuple[int, dict[str, list]]:
@@ -316,7 +337,6 @@ class BenchHelper():
         Returns `(err, result)`, where a non-zero value for `err` describes that the
         output did not match the expected format.
         """
-
         result = { "devices": [] }
         table_regex = None
 
@@ -326,6 +346,8 @@ class BenchHelper():
             table_regex = r"\s*(?P<name>.+?)\s*?:\s+(?P<iops>[0-9.]+)\s+(?P<mibs>[0-9.]+)\s+(?P<avg_lat>[0-9.]+)\s+(?P<min_lat>[0-9.]+)\s+(?P<max_lat>[0-9.]+)"
         elif self.tool in ["xnvmeperf", "xnvmeperf-cuda"]:
             table_regex = r"\s*(?P<name>\w+):?\s+(?P<cpus>[0-9,]+)?\s+(?P<iops>[0-9.]+)\s+(?P<mibs>[0-9.]+)\s+(?P<fails>[0-9.]+)"
+        elif self.tool == "fio_xnvme":
+            return self._parse_fio_results(table)
         else:
             log.error(f"Unkown tool: {self.tool}")
             return -1, None
@@ -341,6 +363,47 @@ class BenchHelper():
 
         return 0, result
 
+    def _parse_fio_results(self, output: str) -> Tuple[int, dict[str, list]]:
+        start = output.find("{")
+        end = output.rfind("}")
+        if start < 0 or end < 0 or end <= start:
+            log.error("Failed: could not find fio JSON output")
+            return 1, None
+
+        try:
+            payload = json.loads(output[start:end + 1])
+        except json.JSONDecodeError:
+            log.error("Failed: invalid fio JSON output")
+            return 1, None
+
+        jobs = payload.get("jobs", [])
+        if not jobs:
+            log.error("Failed: fio returned no jobs")
+            return 1, None
+
+        total_iops = 0.0
+        total_mibs = 0.0
+        devices = []
+        for job in jobs:
+            metrics = job.get("read", {})
+            if not float(metrics.get("iops", 0.0)):
+                write_metrics = job.get("write", {})
+                if float(write_metrics.get("iops", 0.0)):
+                    metrics = write_metrics
+            iops = float(metrics.get("iops", 0.0))
+            mibs = float(metrics.get("bw_bytes", 0.0)) / (1024 * 1024)
+            total_iops += iops
+            total_mibs += mibs
+            devices.append({"iops": iops, "mibs": mibs})
+
+        return 0, {
+            "devices": devices,
+            "total": {
+                "iops": total_iops,
+                "mibs": total_mibs,
+            },
+        }
+
     def _parse_time_output(self, output: str) -> Tuple[int, int]:
         """
         Find the CPU usage from /usr/bin/time in from the output of /usr/bin/time.
@@ -348,7 +411,6 @@ class BenchHelper():
         Returns `(err, cpu_usage)`, where a non-zero value for `err` describes that the
         output did not match the expected format.
         """
-
         time_regex = r"(?P<user>[0-9.]+)user (?P<system>[0-9.]+)system (?P<elapsed>[0-9.:]+)elapsed (?P<cpu>[0-9.]+)%CPU .*k"
         m = search(time_regex, output)
 
--- a/scripts/bench_helper.py
+++ b/scripts/bench_helper.py
@@ -251,7 +251,10 @@ class BenchHelper():
                 log.error("Failed: DcgmHelper.stop_and_parse()")
                 return err, None
 
-        cpu_freqs = [[cpu_freqs[idx], self.cpu_pairs[idx]] for idx in selected_cpus]
+        if self.cfm.cpu_control_supported and cpu_freqs:
+            cpu_freqs = [[cpu_freqs[idx], self.cpu_pairs[idx]] for idx in selected_cpus]
+        else:
+            cpu_freqs = []
 
         result = {
             "rw": rw,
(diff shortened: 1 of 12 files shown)


tasks/bench_fio_compare.yaml
```diff
@@ -0,0 +1,219 @@
...
+    results_dir: "{{ local.env.PWD }}/{{ config.fio_compare.output_root }}/{{ config.devices[0].pci_addr }}/raw/read"
+    tool: fio_xnvme
+    backend: spdk
+
+- name: prepare_read_upcie
+  run: |
+    {{ config.xnvme.driver.prefix }} xnvme-driver reset || true
+    bash -lc 'ns_path=$(find /sys/bus/pci/devices/{{ config.devices[0].pci_addr }}/nvme -mindepth 2 -maxdepth 2 -type d -name "nvme*n*" 2>/dev/null | head -n 1); if [ -n "$ns_path" ]; then blkdiscard -f "/dev/$(basename "$ns_path")" || true; fi'
+    modprobe uio_pci_generic
+    DRIVER_OVERRIDE=uio_pci_generic {{ config.xnvme.driver.prefix }} xnvme-driver
```
karlowich: Is there any reason not to do all the SPDK benchmarks first and then all the uPCIe? This way you only need the "prepare" steps once per backend

karlowich (commented):
There's a lot take in here, but it looks very promising!
So far, I have one comment :)
I will do another pass later.

naddinadja:
In regards to the first commit: Why is it necessary to account for cpupower and turbo not being available on the platform? If the machine has been provisioned following the steps and tasks in the AiSIO repo, these should be available, no?

scripts/bench_helper.py
```diff
@@ -131,7 +142,7 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
                 return 0, result
```
naddinadja: Add an `elif self.tool == "fio_xnvme"` that sets the filename, since fio_size is not relevant for any other tools.

scripts/bench_helper.py
```diff
@@ -234,19 +257,23 @@ def abort_monitors():
...
             "cpu_freqs": cpu_freqs,
             "fixed_freq": self.cfm.fixed_freq,
             "cpu_governor": self.cfm.governor,
-            "thr_sib": 1 if self.use_thrsib else 0,
+            "cpu_control_supported": self.cfm.cpu_control_supported,
+            "thr_sib": self.use_thrsib,
             "smt": 1 if "SMT1" in suffix else 0,
             "turbo": 1 if "turbo1" in suffix else 0,
-            "stress": 1 if self.stress else 0,
+            "stress": self.stress,
```
naddinadja: Please keep using 0 and 1 for booleans - it helps when different languages spell the booleans differently (e.g. `true` vs `True`)

scripts/bench_helper.py
```diff
@@ -141,7 +152,23 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
+        if is_cuda:
+            selected_cpus = []
+            command = f"/usr/bin/time "
+        else:
+            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
+            if self.tool == "fio_xnvme" and rw in ["read", "randread"]:
+                err = self._prepare_fio_read_working_set(self.devices[0]["pci_addr"], rw, self.backend)
+                if err:
+                    log.error("Failed: _prepare_fio_read_working_set()")
+                    return err, None
```
naddinadja: If you need to prefill the devices, I think it's better to do this before you run the benchmarking. No need to prefill on every read in the workflow, and I also don't think this belongs inside of the bench helper :)

scripts/fio_xnvme.py
```diff
@@ -0,0 +1,50 @@
+def sh_quote(value: str) -> str:
+    return "'" + value.replace("'", "'\"'\"'") + "'"
```
naddinadja: Can you give an example of what this function is solving? I don't see where there are ticks in the input, but as I read this function, it replaces each normal tick `'` with `'"'"'` and then puts additional ticks `'` around the whole thing?

scripts/fio_xnvme.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        config += [
+            f"io_size={fio_size}",
+        ]
+
+    return "\n".join(config + ["", "[job0]"])
+
+
+def fio_xnvme_cmd(bin: str, fio_job: str) -> str:
+    return f"printf %s {sh_quote(fio_job)} | {bin} - --output-format=json"
```
naddinadja: For simplicity's sake, I would suggest that use CLI arguments instead of trying to construct a fio config file. Then the setup of this file might also match the other benchmark tool files.

scripts/bench_helper.py
```diff
@@ -141,7 +152,23 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
+
+            cpu_list = ",".join(str(cpu) for cpu in selected_cpus)
+            command = f"taskset -c {cpu_list} /usr/bin/time "
```
naddinadja: if taskset is only needed for fio, I suggest we put it in an if-statement

scripts/bench_helper.py
```diff
@@ -150,28 +177,24 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
             bench_args["backend"] = self.backend
             bench_args["nqueues"] = nqueues
             command += xnvmeperf_cuda_cmd(self.bin, bench_args)
-
+        elif self.tool == "fio_xnvme":
+            fio_job = create_fio_job(
+                self.devices[0]["pci_addr"], size, depth, time, rw, self.backend, self.fio_size
+            )
+            command += fio_xnvme_cmd(self.bin, fio_job)
         else:
```
naddinadja, line 191: Did your formatter remove the spaces between the if/elif statements? Might just be my own preference, but I feel like it's more easily read with the spaces in Python, when the if-elif-elif-elif-else statement gets very long :)

naddinadja (changes requested):
I have a couple questions / suggestions. Haven't looked through it all yet, but wanted to give this part of the feedback before going on weekend :)

scripts/bench_helper.py
yonggilsong: Good point. I'll split out the fio_xnvme filename handling into its own branch so fio_size/rw-specific naming stays local to that tool.

scripts/bench_helper.py
yonggilsong: Makes sense. I'll keep this field as 0/1 to stay consistent with the rest of the benchmark result output.

scripts/bench_helper.py
yonggilsong: Agreed. I'll scope taskset to the fio_xnvme path instead of applying it unconditionally.

scripts/bench_helper.py
yonggilsong: Sure, I'll restore the spacing there to keep the long if/elif chain easier to read.

scripts/fio_xnvme.py
yonggilsong: This was added to safely pass the generated fio job text through printf into the shell pipeline without breaking on embedded quotes/newlines. That said, I agree the current approach is not very obvious, and I’ll either document it clearly or remove the need for it as part of the fio_xnvme cleanup.

scripts/bench_helper.py
yonggilsong: I agree with the separation-of-responsibility concern. I’m planning to move this out of BenchHelper and make it explicit in the workflow. While doing that, I’ll also rework the workflow to run all SPDK cases first and all uPCIe cases afterwards, so the prefill/prepare flow can be handled more cleanly there.

tasks/bench_fio_compare.yaml
yonggilsong: I agree the backend-first structure is cleaner and would avoid repeated prepare steps. I'll rework the workflow so all SPDK runs happen together and all uPCIe runs happen together, unless I hit an ordering dependency while moving the read-prefill flow out of the helper.

scripts/fio_xnvme.py
yonggilsong: That sounds like the better direction. I initially kept the fio job text because it was the quickest way to get the compare path working, but I agree CLI arguments would align better with the other benchmark helpers. I'll rework fio_xnvme in that direction.

yonggilsong:
> In regards to the first commit: Why is it necessary to account for cpupower and turbo not being available on the platform? If the machine has been provisioned following the steps and tasks in the AiSIO repo, these should be available, no?

In the ideal setup, yes, these controls should be available after following the AiSIO provisioning flow. The issue I ran into was that this assumption does not always hold in practice on all target machines. In particular, custom dmabuf kernels do not necessarily have a matching linux-tools package available for cpupower, and some platforms still do not expose the expected CPU frequency/turbo control interfaces even after provisioning. I added this handling so the benchmark reports that CPU control was unavailable instead of aborting mid-run or silently misrepresenting the environment.

tasks/bench_fio_compare.yaml
yonggilsong: I separated trim from prepare in the updated workflow. That makes the device-state transition explicit, and it also lets write and randwrite start from a freshly trimmed state instead of inheriting the state left by earlier workloads.

karlowich: I suggest taking the trim code out of the workflow file and creating a separate script for it. That way we can simplify the workflow steps visually.
Likewise, we can create a separate script for the pre-fill step.
I want to move towards reducing the complexity of `bench_helper` and `bench_runall`.

karlowich:
@yonggilsong 
>  In particular, custom dmabuf kernels do not necessarily have a matching linux-tools package available for cpupower

This has been addressed, now we install cpupower alongside the custom kernel.
Can do we still need the CPU-freq skip? It would be a nice simplification if we don't need it :)

tasks/bench_fio_compare.yaml
yonggilsong: That makes sense. I’ll move the trim and pre-fill logic out of the workflow.

Do you have a preference between adding separate action modules like `fio_xnvme_trim.py` / `fio_xnvme_prefill.py`, or keeping trim, pre-fill, and fio command construction together in `fio_xnvme.py`?

The separate modules would make the workflow steps explicit and easier to extend independently, while keeping them in `fio_xnvme.py` would keep the compare-benchmark-specific fio/xNVMe logic in one place.

yonggilsong:
> @yonggilsong
> 
> > In particular, custom dmabuf kernels do not necessarily have a matching linux-tools package available for cpupower
> 
> This has been addressed, now we install cpupower alongside the custom kernel. Can do we still need the CPU-freq skip? It would be a nice simplification if we don't need it :)

Thanks for the update. Could you point me to the branch or PR that addresses installing cpupower alongside the custom kernel?

I’d like to test that setup first and confirm CPU frequency control works with the custom dmabuf kernel. Once that’s verified, I’ll remove the CPU-freq skip/fallback path as a follow-up simplification.

karlowich:
@yonggilsong, sorry for the slow reply. It is here: https://github.com/xnvme/aisio/pull/58/changes/8dcf598d4d0a3c96188ef62b8aec8386a4aa0f35

yonggilsong:
@karlowich Thanks, that is helpful. Since PR #58 now installs `cpupower` alongside the custom kernel, I agree this should be re-validated against the updated provisioning path rather than kept based on the older setup issue alone.

What I still want to verify is whether the remaining fallback is only covering `cpupower` availability, or whether it is also covering hosts where turbo / cpufreq control or frequency sampling is still unavailable even after provisioning. In earlier runs, the latter was the bigger issue on some machines.

I’ll re-test this against the updated setup and then narrow the logic accordingly:
- if the fallback is no longer needed, I’ll remove it
- if only part of it is still needed, I’ll keep the minimum host-capability fallback and drop the rest

scripts/bench_helper.py
```diff
@@ -131,17 +143,35 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
+        if self.tool == "fio_xnvme" and ndevs != 1:
+            log.error("Failed: fio_xnvme currently supports exactly 1 device per benchmark point")
+            return 1, None
+
+        if is_cuda:
+            selected_cpus = []
+            command = f"/usr/bin/time "
+        else:
+            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
+            command = f"/usr/bin/time "
```
naddinadja: The `command = f"/usr/bin/time "` is added to both the if and else path (`if is_cuda: ... else: ...`). How about we keep it outside, before the if-statement :)

naddinadja (changes requested):
I only found this small thing here :)

yonggilsong:
@karlowich @naddinadja I've now re-tested the fallback path against the updated provisioning (PR #58) on a physical AMD EPYC host. Summary of what I found:

**1. cpupower availability alone is not the whole story.**
Even with the updated provisioning, I hit two separate issues before CPU frequency control actually worked:
- On a custom kernel, `cpupower` did not resolve to a working binary out of the box (the distro `linux-tools` package only matches official release kernels), so `cpupower frequency-info` failed until the tooling was aligned with the running kernel.
- More importantly, once `cpupower` ran, **no cpufreq driver was active** (`amd_pstate: the _CPC object is not present in SBIOS or ACPI disabled`). All ACPI tables exposed zero `_CPC`/`_PSS`/`_PPC` objects. This is a **BIOS-level** dependency that provisioning cannot fix.

**2. Once the platform actually exposes CPU frequency control, the fallback never triggers.**
After enabling CPPC/P-State in the BIOS, `acpi-cpufreq` attached, `cpupower frequency-info` reported proper frequency steps, and a full `bench_fio_compare` run produced `cpu_control_supported=1` on every result with real `cpu_freqs` samples and a successful governor set — i.e. none of the fallback branches (cpupower / turbo / cpufreq logging) were hit.

**Conclusion:** the fallback is *not* needed on a correctly-provisioned host whose BIOS exposes CPPC. But it *is* still covering a real gap: provisioning + PR #58 guarantee that `cpupower` is installed, but they do not guarantee that the host's BIOS exposes the cpufreq/turbo interfaces. So rather than removing it entirely, I'll keep a minimal host-capability guard (report `cpu_control_supported=0` and continue) and drop the parts that PR #58 now makes redundant. I'll also document that CPPC must be enabled in BIOS for CPU frequency control to be available.

scripts/bench_helper.py
yonggilsong: Good catch. I'll move `command = f"/usr/bin/time "` out to before the if-statement so it's shared, and only prepend the `taskset` prefix on the path that needs it. That also ties in with your taskset comment below — I'll scope `taskset` to the fio path and keep the common `/usr/bin/time` part in one place.

tasks/bench_fio_compare.yaml
karlowich: I think separate modules are preferred. The the `fio_xnvme_trim.py` and `fio_xnvme_prefill.py` modules can use `fio_xnvme.py`

yonggilsong:
### What changed in this push (vs. the previously reviewed revision)

This revision folds in the review feedback and the CPU-frequency fallback re-validation. Grouped by the comments they address:

**Workflow structure — trim/prefill extraction (@karlowich)**
- `trim` and `prefill` are no longer inline `bash -lc '… awk …'` steps. They are standalone actions, `scripts/fio_xnvme_trim.py` and `scripts/fio_xnvme_prefill.py`, invoked via `uses:`. This shrinks `tasks/bench_fio_compare.yaml` and makes the device-state transitions explicit.
- `trim` is now separate from `prepare`, so each write / randwrite workload starts from a freshly trimmed namespace instead of inheriting the previous workload's state.

**bench_helper / bench_runall (@naddinadja)**
- Dropped the `prefill_only` parameter and the prefill branch from `run_benchmark()`; prefill now lives entirely in the workflow action. `bench_runall.py` dropped `--prefill_only` accordingly.
- Scoped `taskset` to the `fio_xnvme` path instead of applying it unconditionally.
- Hoisted the shared `command = "/usr/bin/time "` out of the `is_cuda` if/else so it is no longer duplicated in both branches.

**fio_xnvme command construction (@naddinadja)**
- `fio_xnvme.py` now builds the fio invocation from CLI arguments (`--name`, `--rw`, `--bs`, `--iodepth`, …) instead of constructing a job-config text and piping it through `printf`. This removes the non-obvious `sh_quote` shell-escaping helper and aligns the file with the other benchmark tool helpers.

**CPU-frequency fallback — scoped down after re-validation**
- Removed the dead `get_cpu_frequency_steps()`; its only effect was a guard for `cpupower` *binary* availability, which PR #58 now guarantees by installing `cpupower` alongside the kernel.
- Kept the guards that PR #58 cannot cover — i.e. the host's BIOS not exposing the cpufreq/turbo interfaces: `set_cpu_freq` / `_clear_fixed_cpu_freq`, `toggle_turbo`, and the logger `UNSUPPORTED` / empty-output handling.
- Validated both directions on a physical AMD EPYC 7742: with BIOS CPPC enabled the fallback never triggers (`cpu_control_supported=1`, full run 128/128 result rows); with CPPC disabled the kept guards trigger gracefully (`cpu_control_supported=0`, smoke 8/8, workflow completes with no crash/abort).

**Bug fixes found during validation**
- `fio_xnvme_prefill.py` declared `--fio-size` (hyphen) while cijoe passes the `with:` key as `--fio_size` (underscore), so the prefill step failed with "required: --fio-size". Fixed to `--fio_size`.
- Restored the cpu-freq logger guards (`UNSUPPORTED` / empty / empty-data) and the matching `cpu_control_supported and cpu_freqs` guard in `bench_helper.py` that an interim refactor had dropped (would `IndexError` on a host without cpufreq).

**Tests added**
- `tests/test_cpu_freq_fallback.py` — unsupported / missing-cpufreq fallback paths.
- `tests/test_fio_compare_refactor.py` — trim/prefill extraction and `fio_size` argument mapping.

karlowich (changes requested):
This is shaping up nicely. 

I would like to drop the python tests and the bench_fio_compare_smoke test.
I don't think they are adding much after development is done and I prefer to keep main free of these "development artifacts".

Also can you squash changes into the original commits instead of placing new commits on top. This makes the experience of reviewing better. Since, the "wrong" version appear first when reviewing commit-wise, it can be confusing to read.

yonggilsong:
@karlowich Both addressed in the latest push (force-updated):

**1. Dropped the dev artifacts** — removed all the python tests and the smoke workflow:
`tests/test_fio_compare_artifacts.py`, `tests/test_fio_compare_validation.py`, `tests/test_fio_compare_refactor.py`, `tests/test_cpu_freq_fallback.py`, `tasks/bench_fio_compare_smoke.yaml`, `configs/bench_fio_compare_smoke.toml`.

**2. Squashed the fixes into their original commits** — the branch is now 3 commits, each correct from the start (no "wrong-then-fixed" intermediate when reading commit-wise):
- `feat(bench/compare): add fio xnvme compare workflow` — folds in the trim/prefill extraction, the `fio_size` prefill-arg fix, and the `/usr/bin/time` hoist
- `fix(config): quote PCI_BLACKLIST values`
- `fix(bench/cpu): handle missing cpu freq controls` — kept as its own commit since the host-capability guards are a distinct concern from the compare workflow

scripts/bench_helper.py
```diff
@@ -131,29 +143,45 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
+            selected_cpus = []
+        else:
+            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
+
         command = f"/usr/bin/time "
 
+        if self.tool == "fio_xnvme":
+            cpu_list = ",".join(str(cpu) for cpu in selected_cpus)
+            command = f"taskset -c {cpu_list} {command}"
+
```
karlowich: Can we merge some of these if statements? There's multiple `if self.tool == "fio_xnvme":` in a row. Also perhaps the `if is_cuda` and `if not is_cuda` could become a single if else.

karlowich (changes requested):
Much nicer!
Just a final thing

scripts/bench_helper.py
yonggilsong: Addressed in the latest push.

I introduced `is_fio = self.tool == "fio_xnvme"` near the top of `run_benchmark()` and used that single flag through the rest of the function. That reduces the repeated `self.tool == "fio_xnvme"` checks in `run_benchmark()` from five occurrences down to one, and keeps the `is_cuda` / non-`is_cuda` CPU-selection flow as a single `if/else` pair.

So there should be no behavioral change here, just a smaller and easier-to-read set of conditionals in the compare path.

scripts/bench_helper.py
```diff
@@ -163,15 +192,12 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
             bench_args["nqueues"] = nqueues
             command += xnvmeperf_cuda_cmd(self.bin, bench_args)
 
+        elif is_fio:
+            command += fio_xnvme_cmd(self.bin, bench_args)
```
karlowich, line 190: I meant, can we merge the if statements above into this one.
Then it becomes something like this:
```python
elif is_fio:
    if ndevs != 1:
            log.error("Failed: fio_xnvme currently supports exactly 1 device per benchmark point")
            return 1, None

    bench_args["backend"] = self.backend
    bench_args["rw"] = rw
    bench_args["fio_size"] = self.fio_size
    cpu_list = ",".join(str(cpu) for cpu in selected_cpus)
    command = f"taskset -c {cpu_list} {command}"    
    command += fio_xnvme_cmd(self.bin, bench_args)
```

I'm not sure of the order of the statements, but I hope you can see what I mean :)

scripts/bench_helper.py
```diff
@@ -131,29 +144,45 @@ def run_benchmark(self, depth: int, size: int, ndevs: int, ncpus: int, time: int
...
             bench_args["cpumask"] = self.cpu_masks[ncpus]
 
+        if is_fio and ndevs != 1:
+            log.error("Failed: fio_xnvme currently supports exactly 1 device per benchmark point")
+            return 1, None
+
+        if is_cuda:
+            selected_cpus = []
+        else:
+            selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
```
karlowich, line 158: Let's merge these into the `if not is_cuda` above, so it becomes:
```python
if is_cuda:
    selected_cpus = []
else:
    bench_args["cpumask"] = self.cpu_masks[ncpus]
    selected_cpus = [v[0] for v in self.cpu_pairs if int(bench_args["cpumask"], 16) & (1 << v[0])]
```

karlowich (changes requested):

scripts/bench_helper.py
yonggilsong: Sorry, I did not fully understand the intended structure from the previous comment. Your follow-up example made it much clearer.

I have now updated it in that direction: the `cpumask` assignment is merged into the non-`is_cuda` branch, and the fio-specific preparation/validation/command construction is grouped more tightly under the `is_fio` path.

Thanks for the concrete guidance — that helped a lot.

karlowich (approved):
Great!

naddinadja:
Your last commit is not signed off, but otherwise it looks good :)

yonggilsong:
> Your last commit is not signed off, but otherwise it looks good :)

Oh, I didn't catch that. I've pushed the signed commit now. Thanks!

naddinadja (approved):
> Oh, I didn't catch that. I've pushed the signed commit now. Thanks!

Yees it's such a small thing haha - but great! Approved from my side too :)

PR #65, opened 2026-05-12
IPC for xal extent tree

naddinadja:
A larger PR with some minor fixes here and there, and then the full functionality for sharing the xal pools with the client. It depends on some of the functionalities from this PR in xal: https://github.com/xnvme/xal/pull/28

### DIFF
--- a/homi/include/homi_proto.h
+++ b/homi/include/homi_proto.h
@@ -3,15 +3,27 @@
 
 #include <stdint.h>
 
+#include <libxal.h>
+
 #define HOMI_MAX_CONNECTS   8
 
 enum homi_msg_type {
-	/* TODO: add message types as daemon functionality is extended */
 	HOMI_MSG_TYPE_HELLOWORLD = 0, ///< Test type, as an example of what is needed
+	HOMI_MSG_TYPE_XAL_CONNECT = 1, ///< Request xal pool info for a device
 };
 
 struct homi_req_helloworld {
 	int32_t value;
+};
+
+struct homi_req_xal_connect {
+	char dev_uri[256];
+};
+
+struct homi_res_xal_connect {
+	int err;
+	struct xal_sb sb;
+	char shm_name[64];
 };
 
 struct homi_msg_header {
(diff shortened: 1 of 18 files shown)


safl (commented):
Does this replace #35?
Should we give homi a repos of its own? E.g. xnvme/homi?

naddinadja:
> Does this replace #35? 

hehe yes, I will close that one

> Should we give homi a repos of its own? E.g. xnvme/homi?

Yes, that would probably be a good idea

homi/src/homic.c
siu-jung, line 1: Is there an advantage to reconnecting to the socket at every request vs. keeping it open?

siu-jung (approved):
lgtm! Nice work :)

homi/include/homi_proto.h
```diff
@@ -3,17 +3,29 @@
 
 #include <stdint.h>
 
+#include <libxal.h>
+
 #define HOMI_MAX_CONNECTS   8
 
 enum homi_msg_type {
-	/* TODO: add message types as daemon functionality is extended */
 	HOMI_MSG_TYPE_HELLOWORLD = 0, ///< Test type, as an example of what is needed
```
karlowich: Perhaps you can also remove the HELLOWORLD, now that we have a "real" message.

homi/src/homic.c
```diff
@@ -98,7 +120,7 @@ homic_helloworld(int32_t value, char **out)
 		return err;
 	}
 
-	err = homi_proto_socket_read(g_homic_client->fd, &hdr, &response);
+	err = homi_proto_socket_read(g_homic_client->fd, &hdr, (void **)&response);
```
karlowich: I don't think you need the `void **` cast

karlowich (changes requested):
Very nice!
Just a few nitpicks

homi/src/homic.c
naddinadja: Not really, but the problem was that it wasn't being kept open, so if I did two different calls, it failed the second one due to the socket having been closed for some reason

naddinadja:
@karlowich I've addressed your comments in the last commit - and also added two commits for docs

naddinadja:
I will close this PR, as we have moved HOMI to a new repository and opened it there: https://github.com/xnvme/homi/pull/1

PR #66, opened 2026-05-13
feat(bench/compare): add uPCIe IOMMU overhead benchmark

jaeyoon-choi:
## Summary

Add an AiSIO CIJOE benchmark for measuring uPCIe IOMMU overhead.

The task automates the full UIO/VFIO comparison flow by switching the target
between IOMMU-off and IOMMU-on boots, collecting benchmark results in each boot
mode, and generating an HTML report from the paired results.

## Changes

- add an IOMMU overhead benchmark config
- add a CIJOE task for the full rebooted benchmark flow
- configure and verify IOMMU boot mode through grub
- collect IOMMU-off `uio_pci_generic` results
- collect IOMMU-on `vfio-pci` results
- collect throughput with `xnvmeperf`
- collect fio throughput, mean latency, and tail latency
- combine UIO/VFIO result pairs and calculate deltas
- generate an HTML report through `bench_visualize.py`

## Usage

```bash
cijoe --monitor \
  -c configs/transport.toml \
  -c configs/aisio.toml \
  -c configs/datasets.toml \
  -c configs/iommu_overhead.toml \
  tasks/bench_iommu_overhead.yaml
```

The workflow updates /etc/default/grub on the target and reboots twice:

1. IOMMU off: collect uio_pci_generic
2. IOMMU on: collect vfio-pci
3. Combine results and render artifacts/benchmark-results.html

## Notes
- The target NVMe BDF is read from filesystems.dset.pci_addr.
- The original grub config is backed up on the target as
/etc/default/grub.aisio-iommu-overhead.bak.
- If recovery is needed, restore that file, regenerate grub config, and reboot.
- The workflow leaves the target in the final IOMMU-on boot mode.

## Result

[benchmark-results.html](https://github.com/user-attachments/files/27691984/benchmark-results.html)

### DIFF
--- a/scripts/iommu_boot.py
+++ b/scripts/iommu_boot.py
@@ -1,0 +1,157 @@
+"""
+Configure and verify IOMMU boot mode
+====================================
+
+This script edits `/etc/default/grub` on the CIJOE target, regenerates the
+boot loader config, and verifies the active boot mode after reboot.
+"""
+
+import errno
+import logging as log
+import shlex
+from argparse import ArgumentParser
+from pathlib import Path
+
+from cijoe.core.resources import get_resources
+
+from iommu_common import cmdline_has_iommu_off, dmesg_indicates_iommu_enabled
+
+GRUB_UPDATE_REMOTE = "/tmp/aisio-iommu-grub-update.py"
+GRUB_UPDATE_RESOURCE = "iommu_grub_update"
+
+
+def add_args(parser: ArgumentParser):
+    parser.add_argument(
+        "--mode",
+        choices=["set-off", "set-on", "verify-off", "verify-on"],
+        required=True,
+    )
+
+
+def quote_shell_arg(value):
+    """Return a shell-quoted string for remote command construction."""
+    return shlex.quote(str(value))
+
+
+def artifacts_path(args):
+    path = Path(args.output) / "artifacts" / "iommu-overhead"
+    path.mkdir(parents=True, exist_ok=True)
+    return path
+
+
+def read_target_file(cijoe, path):
+    err, state = cijoe.run(f"cat {quote_shell_arg(path)}")
+    if err:
+        return err, None
+    return 0, state.output()
+
+
+def detect_grub_update_command(cijoe):
+    """Return the distro-appropriate command to regenerate grub config."""
+    err, _ = cijoe.run("command -v update-grub")
+    if not err:
+        return "update-grub"
+    err, _ = cijoe.run("command -v grub2-mkconfig")
+    if not err:
+        return "grub2-mkconfig -o /boot/grub2/grub.cfg"
+    return None
+
+
+def upload_grub_update_script(cijoe):
+    if not (
+        script := get_resources().get("auxiliary", {}).get(GRUB_UPDATE_RESOURCE, {})
+    ):
+        log.error("Failed retrieving the grub update script from auxiliary files")
+        return errno.ENOENT
+
+    cijoe.run(f"rm -f {quote_shell_arg(GRUB_UPDATE_REMOTE)}")
+
+    if not cijoe.put(script.path, GRUB_UPDATE_REMOTE):
+        log.error("Failed transferring grub update script from initiator to target")
+        return errno.EIO
+
+    return 0
+
+
+def set_mode(args, cijoe, mode):
+    artifacts = artifacts_path(args)
+    err, before = read_target_file(cijoe, "/etc/default/grub")
+    if err:
+        log.error(f"Failed reading /etc/default/grub: {err}")
+        return err
+    if before is not None:
+        (artifacts / f"grub-before-{mode}.txt").write_text(before)
+
+    grub_update = detect_grub_update_command(cijoe)
+    if grub_update is None:
+        log.error("No update-grub or grub2-mkconfig found on target")
+        return errno.ENOENT
+
+    err = upload_grub_update_script(cijoe)
+    if err:
+        log.error("Failed transferring grub update script")
+        return err
+
+    cmd = (
+        f"python3 {quote_shell_arg(GRUB_UPDATE_REMOTE)} "
+        f"{quote_shell_arg(mode)} && {grub_update}"
+    )
+    err, state = cijoe.run(cmd)
+    (artifacts / f"update-grub-{mode}.txt").write_text(state.output())
+    if err:
+        log.error(f"Failed updating grub for IOMMU {mode}: {state}")
+        return err
+
+    err, after = read_target_file(cijoe, "/etc/default/grub")
+    if err:
+        log.error(f"Failed reading /etc/default/grub: {err}")
+        return err
+    if after is not None:
+        (artifacts / f"grub-after-{mode}.txt").write_text(after)
+
+    return 0
+
+
+def verify_mode(args, cijoe, mode):
+    artifacts = artifacts_path(args)
+
+    err, cmdline_state = cijoe.run("cat /proc/cmdline")
+    if err:
+        log.error(f"Failed reading /proc/cmdline: {cmdline_state}")
+        return err
+    cmdline = cmdline_state.output()
+
+    err, dmesg_state = cijoe.run("dmesg | grep -i -E 'DMAR|IOMMU|AMD-Vi' || true")
+    if err:
+        log.error(f"Failed reading dmesg: {dmesg_state}")
+        return err
+    dmesg = dmesg_state.output()
+
+    (artifacts / f"iommu-verify-{mode}.txt").write_text(
+        f"=== /proc/cmdline ===\n{cmdline}\n=== dmesg ===\n{dmesg}"
+    )
+
+    off_in_cmdline = cmdline_has_iommu_off(cmdline)
+    enabled_in_dmesg = dmesg_indicates_iommu_enabled(dmesg)
+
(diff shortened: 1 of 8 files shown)


scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+from pathlib import Path
+
+from iommu_common import cmdline_has_iommu_off, dmesg_indicates_iommu_enabled
+
+GRUB_UPDATE_REMOTE = "/tmp/aisio-iommu-grub-update.py"
+GRUB_UPDATE_SCRIPT = r"""
+import re
+import sys
+from pathlib import Path
+
```
karlowich: Can we avoid having this python script as a string?
Perhaps we can do all the operations through ssh with CIJOE?
I'm unsure which is easier to read and maintain, I'm not very fond of complicated `sed` commands either.

scripts/iommu_overhead.py
```diff
@@ -0,0 +1,309 @@
...
+def conf(cijoe, key, default=None):
+    return cijoe.getconf(f"iommu_overhead.{key}", default)
+
+
+def cpu_to_cpumask(cpu):
+    return hex(1 << int(cpu))
+
+
+def expected_iommu_enabled(driver):
+    return driver == "vfio-pci"
```
karlowich: Since this helper is only used in one place, I think we can just inline it.

karlowich (changes requested):
Very nice. I found just a few things :)

scripts/iommu_boot.py
naddinadja: A suggestion: Have the python script, or shell if easier, as an auxilliary file, transfer it and run it on the remote. This is what I do with the CPU frequency logger e.g., so we cantake inspiration from there? :)

configs/iommu_overhead.toml
naddinadja, line 1: Is there a reason we don't define these in the tasks? I'm not opposed to this, but I'm just curious since we do it the other way for all other tasks

scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+grub = Path("/etc/default/grub")
+backup = Path("/etc/default/grub.aisio-iommu-overhead.bak")
+text = grub.read_text()
+if not backup.exists():
+    backup.write_text(text)
+
+cpuinfo = Path("/proc/cpuinfo").read_text(errors="ignore").lower()
+vendor = "amd" if "authenticamd" in cpuinfo else "intel"
+off_token = "amd_iommu=off" if vendor == "amd" else "intel_iommu=off"
+on_token = "amd_iommu=on" if vendor == "amd" else "intel_iommu=on"
```
naddinadja: You can simplify this part to just be:
```python
token = f"{vendor}_iommu={mode}"
```
And then you don't have to do if/else in line 51, and in line 62, you don't have to redefine `token` :)

scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+def add_args(parser: ArgumentParser):
+    parser.add_argument(
+        "--mode",
+        choices=["set-off", "set-on", "verify-off", "verify-on"],
+        required=True,
+    )
+
+
+def q(value):
+    return shlex.quote(str(value))
```
naddinadja: Can we add a doc-string here? I haven't worked with shlex before, and the name `q()` doesn't really tell what is going on - but I understand the short name for keeping lines shorter further down. However, if all it does is put qutation marks around the given `value`, you could also just put the quotation marks in the two f-strings where you call `q` :)

scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+    path = Path(args.output) / "artifacts" / "iommu-overhead"
+    path.mkdir(parents=True, exist_ok=True)
+    return path
+
+
+def read_target_file(cijoe, path):
+    err, state = cijoe.run(f"cat {q(path)}")
+    if err:
+        return None, err
+    return state.output(), 0
```
naddinadja: This is not wrong, we just usually put the error first and then the result, hehe

scripts/iommu_boot.py
karlowich: That works for me

scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+    err, cmdline_state = cijoe.run("cat /proc/cmdline")
+    if err:
+        log.error(f"Failed reading /proc/cmdline: {cmdline_state}")
+        return err
+    cmdline = cmdline_state.output()
+
+    err, dmesg_state = cijoe.run("dmesg | grep -i -E 'DMAR|IOMMU|AMD-Vi' || true")
+    if err:
+        log.error(f"Failed reading dmesg: {dmesg_state}")
+        return err
```
naddinadja, line 127: when you say `(...) || true` in the command, does it ever return an error?

scripts/iommu_boot.py
```diff
@@ -0,0 +1,202 @@
...
+        if off_in_cmdline:
+            log.error(
+                "Expected IOMMU-on boot, but /proc/cmdline still has *_iommu=off token"
+            )
+            return errno.EINVAL
+        if not enabled_in_dmesg:
+            log.error(
+                "Expected IOMMU-on boot, but dmesg does not indicate IOMMU is enabled"
+            )
+            return errno.EINVAL
```
naddinadja: You can simplify this by comparing the booleans:
```python
if (mode == "off") != off_in_cmdline:
    log.error(f"Expected IOMMU-{mode}, but /proc/cmdline shows the opposite")
if (mode == "on") != enabled_in_dmesg:
    log.error(f"Expected IOMMU-{mode}, but dmesg indicate the opposite")
```

naddinadja (changes requested):
This is awesome so far, I just have a couple things inline :)

scripts/iommu_overhead.py
jaeyoon-choi: Done!

scripts/iommu_boot.py
jaeyoon-choi: I applied Nadja's suggestion. It looks much better!

configs/iommu_overhead.toml
jaeyoon-choi: Done!

scripts/iommu_boot.py
jaeyoon-choi: Done!

scripts/iommu_boot.py
jaeyoon-choi: Good point. I renamed it to quote_shell_arg() and added doc-string.

scripts/iommu_boot.py
jaeyoon-choi: Thank you!

scripts/iommu_boot.py
jaeyoon-choi: `err` is the exit code, and `dmesg_state` holds the command output. `|| true` keeps a missing grep match from failing the command.

scripts/iommu_boot.py
jaeyoon-choi: Done. Thanks!

scripts/iommu_boot.py
karlowich: Great!

karlowich (approved):
Nice!

scripts/iommu_overhead.py
```diff
@@ -0,0 +1,329 @@
...
+    if err:
+        return err
+
+    total = len(cases) * repeat * 2
+    done = 0
+    try:
+        for case_idx, (rw, iosize, iodepth) in enumerate(cases, start=1):
+            for rep in range(1, repeat + 1):
+                workload = (
+                    f"{args.label} {rw} iosize={iosize} " f"iodepth={iodepth} rep={rep}"
```
naddinadja: I assume this was on two lines, but that the formatter has evaluated that it can fit on one, so maybe change it to be only one string:
```python
f"{args.label} {rw} iosize={iosize} iodepth={iodepth} rep={rep}"
```

scripts/iommu_overhead.py
```diff
@@ -0,0 +1,329 @@
...
+                write_result(path, result)
+                done += 1
+
+            if workload_pause > 0 and case_idx < len(cases):
+                time.sleep(workload_pause)
+
+        for case_idx, (rw, iosize, iodepth) in enumerate(cases, start=1):
+            for rep in range(1, repeat + 1):
+                workload = (
+                    f"{args.label} {rw} iosize={iosize} " f"iodepth={iodepth} rep={rep}"
```
naddinadja: Same as above :)
I assume this was on two lines, but that the formatter has evaluated that it can fit on one, so maybe change it to be only one string:
```python
f"{args.label} {rw} iosize={iosize} iodepth={iodepth} rep={rep}"
```

naddinadja (changes requested):
This is great, there's only this one minor thing, but otherwise it's good to go :)

scripts/iommu_overhead.py
jaeyoon-choi: Done. Thanks!

scripts/iommu_overhead.py
jaeyoon-choi: Done. Thanks!

naddinadja (approved):
Great! Looks good to me :)

PR #70, opened 2026-06-30
Repair docs build and retire the GDS proof-of-concept tooling

safl:
Unbreaks the Sphinx build that fce026b left reading a deleted nvstack.toml, rebuilds the cuda-samples p2p reference at a fixed path so it needs no config knob, and purges the gdsio/GDS-only benchmark tooling (kept in the white-paper as poc-tag references).


karlowich (approved):
LGTM

PR #71, opened 2026-06-30
fix(tasks): install the patched linux-libc-dev for the udmabuf UAPI

safl:
The bindeb-pkg build produces a linux-libc-dev with the patched userspace udmabuf.h (UDMABUF_ATTACH/GET_MAP), but the install step only took linux-headers + linux-image, so /usr/include/linux/udmabuf.h stayed the distro mainline and user-space dma-buf importers fell back to the -ENOTSUP stub. Install it too so the importer ioctls are available without vendoring the UAPI.


karlowich (approved):
LGTM

PR #72, opened 2026-07-01
feat(aisio): bump XAL to v0.3.0 and FIL to v0.3.1

karlowich:
## Summary

Upgrades the AiSIO stack to xal v0.3.0 / fil v0.3.1 and expands the benchmark
suite to exercise the new CPU, GPU, and peer-to-peer I/O paths across all three
datasets.

## Toolchain

- Bump xal to v0.3.0 and fil to v0.3.1.
- xal v0.3.0 relies on BPF, so the build now needs a BTF-enabled kernel and
  `bpftool`:
  - Kernel config switches from a debug-info-free build to DWARF4 + `DEBUG_INFO_BTF`,
    with the resulting `-dbg` image package filtered out at install time.
  - `bpftool` is built and installed from the kernel source tree.
  - `clang`, `llvm`, and libbpf/libelf/zlib dev packages are added to the xal
    and kernel build steps.

## Benchmarks

- `filperf.py` gains explicit per-backend handling for `cufile` and the new
  `aisio-cpu` / `aisio-gpu` / `aisio-p2p` backends, plus `--warmup`,
  `--queue-depth`, and `--max-file-size` options, and emits `--summary`.
- `bench_aisio` now runs CPU, GPU, and P2P variants for `tiktokish`,
  `imagenetish`, and `filesize8gib`. The `filesize8gib` dataset is no longer
  skipped — capping per-file size lets it fit within the buffer budget.
- README bumps the hugepage count to 5120 (10 GiB) so the largest batch of I/O
  buffers fits, with a note on why.

## Cleanup

- Drop the unused `gds` subdirectory from dataset setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,8 +146,12 @@ the dataset volume, rebind the device, and allocate hugepages:
 modprobe uio_pci_generic
 umount /mnt/datasets
 devbind --device '0000:01:00.0' --bind uio_pci_generic  # pci_addr from datasets.toml
-hugepages setup --count 1024
+hugepages setup --count 5120
 ```
+
+The count must be large enough to fit the biggest batch of I/O buffers. The
+`filesize8gib` dataset drives this requirement — 5120 pages (10 GiB) is
+sufficient for the benchmarks in this suite.
 
 `devbind` and `hugepages` are installed on the target as part of `setup_aisio.yaml`.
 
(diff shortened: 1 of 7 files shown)


safl (commented):
A minor thing, in the docs, i recall that there are mentions of the versions of the various components, that should be updated as well.

karlowich:
> A minor thing, in the docs, i recall that there are mentions of the versions of the various components, that should be updated as well.

They're dynamic, so it is all controlled by what is in the config.

safl (approved):
lgtm!

PR #76, opened 2026-08-03
Expand GPU metrics collection for device-initiated I/O benchmarks

yonggilsong:
## Summary

- **feat(dcgm)**: expand the DCGM fields collected during `bench_cuda_*` runs beyond PCIe RX (1010) — GR/SM activity, SM occupancy, DRAM activity, PCIe TX/RX, clocks, and throttle/replay guard fields. Results are stored per-field as `{mean, p95, min, max}`; `bench_combine` aggregates the new dict schema while remaining compatible with legacy scalar results.
- **feat(tasks/setup)**: provision DCGM on targets in the setup workflows and add a validation script for the collection path.
- **fix(plots,templates)**: chart data-filter and rendering fixes (pre-existing issues surfaced while extending the reports).
- **feat(plots)**: plot GPU engine activity for the CUDA benchmarks and annotate the PCIe saturation barplot with the measured link spec (fields 237/238); experiment docs updated accordingly.

## Validation

- Collection path validated on a local GPU (DCGM 4.6.0) with the included validation script, and combine/plot consumers exercised against both the new and legacy result schemas.
- DCGM provisioning steps verified end-to-end on a fresh Ubuntu guest.


scripts/validate_dcgm_helper.py
naddinadja, line 1: Can we build it as a normal cijoe script with a `main(args, cijoe)`? :) You should be able to run single cijoe-compatible scripts with `cijoe path/to/script.py <... other args>`. Then you can just place it under the `scripts` directory and skip line 27-28 where the path is modified

naddinadja (changes requested):
Can you give some screenshots of your results? It's difficult to comprehend how it looks without running the benchmarks yourself :)

scripts/docs_plots_barplot_sat.py
```diff
@@ -24,7 +24,9 @@
     "stress": 0,
     "turbo": 1,
     "smt": 1,
-    "thr_sib": 0,
+    # bench_helper stores thr_sib as a JSON boolean; jq does not coerce
+    # (false == 0 is false), so the filter value must be a bool too
```
karlowich: We don't need this comment justifying the change.

scripts/docs_plots_barplot_tool.py
```diff
@@ -25,7 +25,9 @@
     "cpu_governor": "performance",
     "turbo": 1,
     "smt": 1,
-    "thr_sib": 1,
+    # bench_helper stores thr_sib as a JSON boolean; jq does not coerce
+    # (true == 1 is false), so the filter value must be a bool too
```
karlowich: We don't need this comment justifying the change.

scripts/docs_plots_lineplot.py
```diff
@@ -23,7 +23,9 @@
     "ndevs": 16,
     "cpu_governor": "performance",
     "stress": 0,
-    "thr_sib": 0,
+    # bench_helper stores thr_sib as a JSON boolean; jq does not coerce
+    # (false == 0 is false), so the filter value must be a bool too
```
karlowich: We don't need this comment justifying the change.

docs/src/experiments/device_initiated_iosize.md
```diff
@@ -121,6 +121,27 @@ in the command path. ``qdepth=1`` still achieves 41.7 GB/s at this I/O size,
...
 
+### GPU Compute Cost of the Polling Kernel
+
+```{figure} /lineplot-cuda-iosize-sm.png
+:alt: GPU engine activity vs. I/O size for xnvmeperf (cuda-run) at qdepth=128
+:width: 700px
+:align: center
+
+GPU engine activity vs. I/O size for xnvmeperf (cuda-run), 16 NVMe devices,
+nqueues=1, queue depth fixed at 128 (2048 CUDA threads). SM active and SM
```
karlowich: The `2048 CUDA threads` looks wrong here, or perhaps `nqueues=1, qd 128` is wrong

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -93,6 +102,25 @@ it saturates the device at 4096 total threads, matching ``nqueues=2`` at
...
+
+GPU engine activity vs. queue depth for xnvmeperf (cuda-run), 16 NVMe devices,
+512 B I/O, nqueues fixed at 1. SM active and SM occupancy (DCGM fields
+1002/1003) measure the compute footprint of the persistent polling kernel;
+DRAM active (1005) tracks the GPU memory bandwidth consumed by the incoming
+P2P writes.
+```
+
+The activity plot holds ``nqueues`` at 1 so queue depth is the only variable:
+the thread count grows from 16 to 8192 along the x-axis, and fields 1002/1003
```
karlowich: This doesn't look correct, do we test this up to `qd 8192`?

karlowich (changes requested):
Will be great to get some more detailed insights for GPU performance!

Couple of comments inline.

scripts/validate_dcgm_helper.py
yonggilsong: Done — moved to `scripts/validate_dcgm_helper.py` and converted to a regular cijoe script with `main(args, cijoe)`, which also removed the `sys.path` manipulation. It now runs as `cijoe -c configs/local_gpu.toml scripts/validate_dcgm_helper.py`.

scripts/docs_plots_barplot_sat.py
yonggilsong: Removed (the rationale stays in the commit message).

scripts/docs_plots_barplot_tool.py
yonggilsong: Removed.

scripts/docs_plots_lineplot.py
yonggilsong: Removed.

docs/src/experiments/device_initiated_iosize.md
yonggilsong: The numbers are consistent — the total thread count is `qdepth × nqueues × ndevs`, so 128 × 1 × 16 devices = 2048 CUDA threads, matching the parameter table at the top of the document. Reworded the caption to spell out the arithmetic so it doesn't read like a per-queue count.

docs/src/experiments/device_initiated_qdepth.md
yonggilsong: 8192 here is the total CUDA thread count, not the queue depth — the sweep tops out at `qdepth=512`, and at `nqueues=1` across 16 devices that is 512 × 16 = 8192 threads (the same arithmetic as the `> 8192` entry in the saturation table above). Reworded the paragraph to state the formula and both endpoints explicitly.

yonggilsong:
Pushed an update addressing the inline comments: the validation script is now a regular cijoe script under `scripts/`, the filter comments are dropped, and the two figure captions now spell out the thread-count arithmetic (the sweeps themselves are unchanged — `qdepth` tops out at 512; 2048 and 8192 are total thread counts, `qdepth × nqueues × ndevs`).

@naddinadja here is how the plots render. One note on provenance: the cuda-run benchmarks need the GPU + 16-NVMe target, so these are rendered through the actual plotting pipeline but from synthetic result files shaped after the numbers in the experiment docs. I will regenerate them from real runs once the target is available.

**PCIe saturation barplot, now annotated with the link spec measured via DCGM fields 237/238** (the dashed roofline is derived from the measured gen/width instead of a hardcoded constant):

barplot-sat

**Existing bandwidth / IOPS lineplots** (unchanged layout, now fed by the per-field dict schema):

lineplot-cuda-iosize

lineplot-cuda-qdepth

**New GPU engine activity plots** (SM active / SM occupancy / DRAM active, DCGM fields 1002/1003/1005) — the GPU-side cost of the persistent polling kernel across the two sweeps:

lineplot-cuda-iosize-sm

lineplot-cuda-qdepth-sm

docs/src/experiments/device_initiated_iosize.md
karlowich: Right of course, I overlooked that part.

docs/src/experiments/device_initiated_qdepth.md
karlowich: Sorry, I just misread the text.

karlowich (approved):
Great!

naddinadja (approved):
Nice! :)

PR #77, opened 2026-08-05
Bump fil to v0.3.3 and copy aisio-cpu reads to GPU

karlowich:
Bumps the fil pin to **v0.3.3** and enables its new `--copy-to-gpu` option
on the aisio-cpu benchmark path. fil v0.3.3 adds an opt-in host-to-device
copy for the aisio-cpu/posix backends; passing it here makes the aisio-cpu
benchmark end with the data GPU-resident (like aisio-p2p/-gpu), so the
host-to-device transfer that the P2P path avoids is included in the
measurement rather than hidden.

The copy is toggleable via `--copy_to_gpu` (defaults to on for aisio-cpu).

## Measured effect (RTX PRO 2000 Blackwell, Samsung 980 PRO)

Isolating the copy on `aisio-cpu`:

| Dataset | No copy | With copy | Overhead |
|---|---|---|---|
| tiktokish | 4456 MiB/s | 2633 MiB/s | −41% |
| filesize8gib | 4471 MiB/s | 2745 MiB/s | −39% |


naddinadja (approved):
goodie :)

safl (approved):
Great!

PR #79, opened 2026-08-06
License AiSIO under BSD 3-Clause

karlowich:
This adds a LICENSE file, SPDX identifiers, and sets up REUSE to ensure that all files are licensed from now on.

@safl You are the author of the only GPL file, I propose you change that license to BSD 3-Clause if possible.


safl (approved):
This is excellent, please just change ``auxiliary/generate_dataset.sh`` to BSD-3 as well.
Thanks!

LICENSES/BSD-3-Clause.txt
```diff
@@ -0,0 +1,11 @@
+Copyright (c) <year> <owner>.
```
naddinadja, line 1: Is this supposed to be filled out or stay as `<year> <owner>`?

naddinadja (commented):
one question

LICENSES/BSD-3-Clause.txt
karlowich: I know it looks weird, but this is the correct version of the license.
We get it from `reuse download --all` which puts all necessary licenses in the repo.

naddinadja (approved):
great :)

PR #80, opened 2026-08-17
Fixes before GPU-related benchmark re-runs

karlowich:
Since #76 is now merged, we need to re-run the affected benchmarks to benefit from the expanded collection of GPU metrics.
In the process of running the benchmarks, I stumbled upon a number of minor issues to fix.
The benchmark re-runs are coming in a separate PR after.


scripts/dcgm_helper.py
```diff
@@ -135,17 +173,41 @@ def stop_and_parse(self) -> Tuple[int, Dict[str, dict]]:
...
+            # anything transferred, so the window is described whole rather
+            # than left empty. active_fraction stays at zero, which is what
+            # marks the statistics as covering an idle window.
+            log.warning("No active samples in the dcgm window; describing all of it")
+            active = list(range(count))
+
         stats = {}
         for field, values in raw.items():
-            if not values:
+            selected = [values[idx] for idx in active if idx < len(values)]
```
safl: I did not have the mental capacity for the changes to this helper, so this is my agent's reading rather than mine, and I would like you to check it rather than take it on faith.

Sample selection is now positional across fields, but the parser at line 174 skips unparsable entries rather than recording a gap. A field that reports `N/A` on any line ends up with a shorter list, so index `i` in that field no longer denotes the same sample instant as index `i` in fields 1001 and 1010, from which `active` was built. The selected window for that field is then shifted by the number of skipped samples, and mean, p95, min and max silently describe the wrong samples. Before this change every value was used, so the skew was harmless; it only becomes a defect now. DCGM profiling fields commonly report `N/A` for the first samples after a metric group is created, which is exactly why that `except ValueError` exists. Appending `None` and dropping it after selection would keep all fields the same length.

scripts/nvbandwidth.py
```diff
@@ -0,0 +1,53 @@
...
+    install_path = cijoe.getconf("nvidia.nvbandwidth.path", "/root/git/nvbandwidth")
+    bin = Path(install_path) / "build" / "nvbandwidth"
+
+    err, state = cijoe.run(f"{bin} -t host_to_device_memcpy_ce")
+    if err:
+        log.error(f"Failed: run(nvbandwidth); err({err})")
+        return err
+
+    # nvbandwidth totals the per-GPU matrix in a "SUM <testcase> <value>" line
+    m = search(
```
safl, line 49: The regex reads the `SUM` line, which nvbandwidth totals across every GPU in its matrix. I understand this fits the environment as it now stands, with a single GPU, but won't it fail on a host with more than one?

On such a host the recorded value would be the sum of all their links, and it is then drawn as the single-link practical ceiling in `barplot-sat` and `lineplot-cuda-iosize`. Since it coincides with the right number on a single-GPU host, that failure would be quiet. Pinning the run with `CUDA_VISIBLE_DEVICES` set to the DCGM GPU, or parsing the matrix cell instead of the SUM, would remove the dependence on host topology.

scripts/version_helper.py
```diff
@@ -0,0 +1,67 @@
...
+
+    trimmed = url.strip().removesuffix(".git").replace(":", "/")
+    parts = [part for part in trimmed.split("/") if part]
+
+    return "/".join(parts[-2:]) if len(parts) >= 2 else trimmed
+
+
+def xnvme_version(cijoe: Cijoe) -> str:
+    """
+    Identify the xNVMe checkout the target's binaries were built from, as
```
safl: The docstring says this identifies the checkout the binaries "were built from", but it reads the target at plot time. With `results_dir` continuation the plot step can run days later, or on a different branch, and stamp a build the data did not come from. Saying "read at plot time" in the docstring and the footnote would make the stamp honest.

Better still, could the benchmark capture the state of the repositories and the software at run time, so the stamp is parsed from the collected output instead? `bench_helper.py:262` already assembles the per-run record on the target and writes it as the `.out` JSON, so a `versions` key there would travel with the data and be read back the way `res["dcgm"]` is today.

safl (commented):
The move to nvbandwidth, the narrowed DCGM window and the build stamp all read cleanly, and the new second y axis renders as intended. Most of what I have is inline, and the one on `dcgm_helper.py` is the one I would look at hardest: I did not have the mental capacity for that helper myself, so it comes from my agent rather than from me reading it.

I have not run any of this against a target, so the sample selection, nvbandwidth and version_helper are unexercised. I rendered the new plot path against synthetic data only.

docs/src/experiments/pcie_saturation.md
```diff
@@ -74,9 +74,10 @@ runs affected by link downtraining (generation drop or lane reduction), and an
 increasing replay counter (202) flags retransmissions that reduce effective
 bandwidth — such runs must be excluded from the comparison.
 
-``p2pBandwidthLatencyTest`` from the CUDA samples suite runs a sustained
-bidirectional P2P bandwidth test between two GPUs. The value recorded is the mean
-of the per-direction bandwidth measured simultaneously in both directions.
+``nvbandwidth`` copies from host memory into GPU memory across the GPU's PCIe
+link, the same link the NVMe devices transfer over under P2P, which measures the
+peak bandwidth that link sustains. The value recorded is its
+``host_to_device_memcpy_ce`` result.
```
safl: The methodology here and the metric tables on both experiment pages now name nvbandwidth, while the results prose still cites the `p2pBandwidthLatencyTest` reference of 56.1 GB/s: `pcie_saturation.md:108` and `:124`, and `device_initiated_iosize.md:85` and `:105`. I take the split to be deliberate, since the re-runs land separately, but as it stands each page contradicts itself. A sentence saying the results predate the switch would carry it until the re-run pull request.

karlowich: Yeah, the split was made on purpose. Good point to write it down. I didn't think of that

scripts/nvbandwidth.py
karlowich: Good point that is an oversight.

scripts/dcgm_helper.py
karlowich: Good point. I will add some robustness.

scripts/version_helper.py
karlowich: Right, that's not ideal. Should be fixed.

karlowich:
Added four commits addressing your feedback. Normally, I would squash the changes, but given the density of Python script I think having them on top is clearer

scripts/nvbandwidth.py
safl: Suggestion for good measure: DCGM GPU IDs follow NVML, which is PCI bus order, while CUDA ordinals default to `CUDA_DEVICE_ORDER=FASTEST_FIRST`. On a mixed-GPU host `CUDA_VISIBLE_DEVICES=0` need not be the device DCGM calls 0, so setting `CUDA_DEVICE_ORDER=PCI_BUS_ID` alongside it would make the two agree.

docs/src/experiments/pcie_saturation.md
```diff
@@ -74,9 +74,10 @@ runs affected by link downtraining (generation drop or lane reduction), and an
 increasing replay counter (202) flags retransmissions that reduce effective
 bandwidth — such runs must be excluded from the comparison.
 
-``p2pBandwidthLatencyTest`` from the CUDA samples suite runs a sustained
-bidirectional P2P bandwidth test between two GPUs. The value recorded is the mean
-of the per-direction bandwidth measured simultaneously in both directions.
+``nvbandwidth`` copies from host memory into GPU memory across the GPU's PCIe
```
safl: Nit: now that the run is pinned, this could name the GPU the way the script docstring does, "the PCIe link of the GPU given by `dcgm.gpu`", rather than "the GPU's PCIe link".

safl (commented):
The four follow-ups all read as intended. I reproduced the dcgm one against a synthetic dmon capture: with the profiling fields reporting N/A for the first samples, the clock statistics came out at 1551 MHz before and 1907 MHz after, so the shift is gone. One wording nit inline.

safl (approved):
All four findings are addressed. The two open threads, the `CUDA_DEVICE_ORDER` suggestion and the wording nit, are for you to take or leave; neither blocks this.

On keeping the four follow-ups on top: that was the right call for the review, and now that they have been read they would be fine to fold into the commits they fix.

scripts/bench_combine.py
```diff
@@ -63,10 +65,12 @@ def main(args, cijoe: Cijoe):
 
         for run in bdev_results.glob(f"{path.stem[:-1]}*"):
             with open(run) as file:
-                repeated_results.append(json_load(file))
+                # Files written before the runs recorded a build stamp hold no
+                # "versions" key, and the merge below wants one key set.
+                repeated_results.append({"versions": {}, **json_load(file)})
```
yonggilsong, line 70: `versions` gets a default here for older `.out` files, but `active_fraction` — the other key this PR adds — does not. `combine_dcgm` takes its field list from `ns[0]` and indexes `run[field]` for the rest, so a `results_dir` mixing both vintages either raises `KeyError` or drops the key, depending on the unsorted glob order on line 66.

Not a regression: `combine_dcgm` is untouched here and `main` already has the same exposure from the field set #76 expanded. So the question is which assumption holds — are mixed-vintage result dirs meant to be supported, in which case `active_fraction` wants the same shim, or is `results_dir` always emptied before a run, in which case neither is needed?

docs/tooling/plots.py
```diff
@@ -277,15 +281,40 @@ def lineplot(artifacts, output, driver, xaxis="ncpus", colormap=None):
...
 
         ax.fill_between(x, data - std, data + std, alpha=0.15, color=color)
         ax.plot(x, data, color=color, linewidth=2, marker="o", markersize=4,
                 label=label)
 
+    ax2 = None
+    if y2groups:
+        ax2 = ax.twinx()
+        ax2.set_ylabel(cfg.get("y2label", ""))
+        ax2.spines["top"].set_visible(False)
```
yonggilsong: `axes.spines.right` is `False` in the global `rcParams` at line 37 and `twinx()` inherits it, so `ax2` draws its ticks and `y2label` against no axis line. Hiding `spines["top"]` here is a no-op for the same reason.

My agent rendered the `-sm` plot and the right-hand axis came out bare. Is that intended, or should `ax2.spines["right"].set_visible(True)` go with it?

scripts/docs_plots_lineplot_cuda_iosize.py
```diff
@@ -126,17 +142,23 @@ def main(args, cijoe):
...
             "results": results,
-            "cuda_bandwidth": cuda_bandwidth,
+            "h2d_bandwidth": h2d_bandwidth,
+            "xnvme_version": version,
         }))
 
     for iosize, metrics in sm_results.items():
         for key, values in metrics.items():
             sm_results[iosize][key] = [round(v, 2) for v in avg_stddev(values)]
+        for key, values in state_results.get(iosize, {}).items():
```
yonggilsong: The clock is merged from inside the loop over `sm_results`, so it is dropped for any sweep point `sm_results` lacks. Field 100 is always available, while 1002/1003/1005 report `N/A` where profiling is unsupported — on such a target `sm_results` is empty, `bars` renders empty, and no figure is produced despite the clock having been collected. Iterating the union would let it stand alone. Same in `docs_plots_lineplot_cuda_qdepth.py:146`.

Is a target without the profiling fields one you expect to hit?

yonggilsong (approved):
A few questions left inline, none of them blockers.

Thanks for following up on what my #76 left short — the narrowed DCGM window especially.

scripts/nvbandwidth.py
karlowich: Good catch

docs/src/experiments/pcie_saturation.md
karlowich: Yes, that's better

scripts/bench_combine.py
karlowich: Ah, that should be fixed. Older runs shouldn't fail because they're missing `active_fraction`.
Good catch.

docs/tooling/plots.py
karlowich: That's an oversight, I will fix that.

scripts/docs_plots_lineplot_cuda_iosize.py
karlowich: I don't expect this to run on machines that don't support the profiling. However, I think adding the robustness is worthwhile.

karlowich:
@safl @yonggilsong I have addressed all your comments in separate commits on top.

Below is a picture of the dual-axis plot that was previously rendered incorrectly.

lineplot-cuda-qdepth-sm

karlowich:
I added another fix on top. I found that the tarballs of new archived runs were missing the `.out` files, so that had to be fixed.

yonggilsong (approved):

PR #81, opened 2026-08-21
Re-run the P2P experiments and report what device-initiated I/O costs the GPU

karlowich:
The published saturation, I/O size and queue-depth archives predate the DCGM
work on `main`. Their statistics averaged the whole monitoring window rather
than the transfer, they carried no engine activity, occupancy or clock fields,
and they recorded neither an `nvbandwidth` reference nor the link the run was
measured on. 

## What this does

Re-collects all three experiments on the current tooling, then follows the
results through the documentation.

The queue-depth experiment gains a second result axis: what the persistent
polling kernel costs the GPU. Its footprint turns out to follow the launch
geometry rather than the work it performs, which separates two configurations
the IOPS curve leaves equivalent and settles which one to choose when the GPU
also has to compute on the data it fetches. The I/O size experiment reports the
same cost across its sweep and can now point at the queue-depth sweep for the
variable that moves it.

Supporting changes to the plotting code, each landing before the figure that
needs it: a log axis, measured values written onto the points, a mark for where
a series reaches a reference, per-field activity figures charted over both
sweep variables, and figure-chosen legend placement.

The methodology sections say what the numbers rest on: the device fill state
every synthetic benchmark assumes and how to reach it, the DCGM statistics as
describing the transfer alone, the clock each activity reading was taken at, and
the reference each run measured for itself. The environment table records the
GPU the server now carries.

## NOTE
The changes to `plots.py` has an influence on multiple graphs. I highly encourage that you build the docs and check them while reviewing. There are too many to include here.

### DIFF
--- a/docs/src/experimental_framework.md
+++ b/docs/src/experimental_framework.md
@@ -92,6 +92,32 @@ devbind --device '<pci_addr>' --bind uio_pci_generic
 hugepages setup --count 1024
 ```
 
+(sec-device-fill-state)=
+#### Device Fill State
+
+Every namespace is formatted before synthetic benchmarks are run, which leaves all
+logical blocks deallocated. Reads of deallocated blocks are served by the
+controller without media access resulting in a synthetically low latency.
+The figures this produces may therefore exceed those seen in real-world,
+data-bearing workloads. They are an upper bound on what the I/O path can drive,
+and are to be read as a measure of system overhead and scalability rather than
+as an indicator of application-level storage performance.
+
+The format is a manual step rather than a workflow step, because it only has to
+be done once: every synthetic benchmark issues ``randread`` exclusively, so none
+of them write to the devices and the state established by the format survives
+across runs. Formatting goes through the kernel NVMe driver, so it precedes the
+binding to ``uio_pci_generic`` above and each device is rebound afterwards:
+
+```
+devbind --device '<pci_addr>' --bind nvme
+nvme format /dev/<ns> --force
+devbind --device '<pci_addr>' --bind uio_pci_generic
+```
+
+
+#### Running the Workflows
+
 The benchmark workflows are parameterised by editing the ``run`` step in
 the respective task file. The keys under ``with`` correspond to
 independent variables. ``numcpus_range`` and ``numdevs_range`` are
--- a/docs/src/experimental_framework.md
+++ b/docs/src/experimental_framework.md
@@ -178,9 +178,15 @@ counters via DCGM and a reference PCIe bandwidth measurement from
 ``nvbandwidth``. Described in detail in
 {ref}`sec-experiments-pcie-bandwidth`.
 
-The ``nvbandwidth`` reference binary is built by
-``setup_nvstack.yaml`` at a fixed path, and DCGM fields default in the
-collector, so no extra config is required.
+``nvbandwidth`` copies from host memory into GPU memory across the PCIe link of
+the GPU given by ``dcgm.gpu``, the same link the NVMe devices transfer over under
+P2P, which measures the peak bandwidth that link sustains. The value recorded is
+its ``host_to_device_memcpy_ce`` result.
+
+This and ``bench_cuda_iosize.yaml``, measure the PCIe bandwidth with
+``nvbandwidth`` as a step of their own run rather than carrying a fixed value
+across experiments. As such, the experiments can quote figures that differ
+slightly in the last digit.
 
 ```
 cijoe --monitor \
@@ -196,9 +202,6 @@ Characterizes the minimum CUDA thread count needed to saturate the PCIe link
 under device-initiated I/O, using **xnvmeperf** with the ``cuda-run`` subcommand
 and the **upcie-cuda** backend, with queue depth as the secondary variable.
 Described in detail in {ref}`sec-experiments-cuda-iosize`.
-
-The ``nvbandwidth`` reference binary is built by
-``setup_nvstack.yaml`` at a fixed path, so no extra config is required.
 
 ```
 cijoe --monitor \
--- a/docs/src/experimental_framework.md
+++ b/docs/src/experimental_framework.md
@@ -115,6 +115,35 @@ nvme format /dev/<ns> --force
 devbind --device '<pci_addr>' --bind uio_pci_generic
 ```
 
+(sec-dcgm-sampling)=
+#### GPU Telemetry
+
+The benchmarks that transfer into GPU memory collect GPU telemetry with
+``dcgmi dmon`` alongside what the benchmark tool itself reports. The fields are
+those of the [DCGM feature
+overview](https://docs.nvidia.com/datacenter/dcgm/latest/user-guide/feature-overview.html),
+which defines each of them; every experiment lists the ones it reports.
+
+Fields are sampled every 100 ms and reported as mean/p95/min/max over the
+samples in which the devices were transferring. A sample counts as transferring
+when the PCIe receive traffic (field 1010) exceeds 100 MB/s or the graphics
+engine (field 1001) is more than 1% active. The monitoring window also spans
+process startup and teardown, whose duration varies with the configuration under
+test, so a mean over the whole window would describe the length of the setup as
+much as the workload. The share of the window that qualified is recorded
+alongside the statistics.
+
+The engine fields report what is running on the GPU: SM activity (1002) how much
+of the device has a warp resident on it, warp slot occupancy (1003) how much of
+the warp capacity of a multiprocessor is taken, and DRAM_ACTIVE (1005) how much
+of the GPU memory bandwidth is in use.
+
+The clocks (fields 100 and 101), the throttle reason bits (112), the PCIe replay
+counter (202) and the link generation and width (237/238) are collected as
+guards on whether runs are comparable: the activity fields are fractions of
+cycles, so a run measured at a sagging clock cannot be held against one measured
+at boost, and a downtrained or retransmitting link changes the bandwidth there
+is to reach.
 
 #### Running the Workflows
 
--- a/docs/src/experiments/device_initiated_qdepth.md
+++ b/docs/src/experiments/device_initiated_qdepth.md
@@ -62,6 +62,7 @@ The namespaces are formatted before the run, as described in
 Instructions for running ``bench_cuda_qdepth.yaml`` are provided in
 {ref}`sec-experimental-framework`.
 
+(sec-experiments-cuda-qdepth-results)=
 ## Results
 
 Results are presented as IOPS vs. queue depth, with one line per number of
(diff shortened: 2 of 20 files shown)


yonggilsong:
Built the docs and checked the figures. One small rendering issue, shown in the attached snapshot.

The titles of the two new qdepth activity figures (lineplot-cuda-qdepth-sm.png and lineplot-cuda-qdepth-occupancy.png) are wider than the figure and get clipped at both edges of the PNG. Shortening the titles or splitting them over two lines would fix it.

Everything else checks out: I recomputed the numbers the sections quote (SM active means, IOPS, saturation depths, transferring shares, DRAM/clock/throttle guards) against the archives and they all match.
issue-title-clipping

karlowich:
Interesting, on my machine the titles are rendered okay. I will look into making a fix to handle wide titles.

lineplot-cuda-qdepth-occupancy
lineplot-cuda-qdepth-sm

karlowich:
@yonggilsong I have pushed a commit addressing this. Now the title should render correctly.
Please verify :)

yonggilsong:
@karlowich I verified on my machine: rebuilt the previously clipped figures with new commit and the titles now fit fully inside the canvas, with no visible loss of legibility. Thanks for the quick fix!

yonggilsong (approved):

docs/src/experimental_framework.md
```diff
@@ -92,6 +92,42 @@ devbind --device '<pci_addr>' --bind uio_pci_generic
...
+#### Device Fill State
+
+Every namespace is formatted before synthetic benchmarks are run, which leaves all
+logical blocks deallocated. Reads of deallocated blocks are served by the
+controller without media access, so this places the devices in a uniform state
+rather than a representative one. A partially written device reaches markedly
+lower random read throughput, and mixing fill states across devices skews the
+aggregate.
+
+The experiments refer to devices left in this state as empty.
```
naddinadja: Can we get quotation marks around empty? I was a bit confused reading this

docs/src/experimental_framework.md
```diff
@@ -92,6 +92,42 @@ devbind --device '<pci_addr>' --bind uio_pci_generic
...
+across runs. Formatting goes through the kernel NVMe driver, so it precedes the
+binding to ``uio_pci_generic`` above and each device is rebound afterwards:
+
+```
+devbind --device '<pci_addr>' --bind nvme
+nvme format /dev/<ns> --force
+devbind --device '<pci_addr>' --bind uio_pci_generic
+```
+
+That uniform state is what these benchmarks are after. With the variability of
```
naddinadja: "That uniform state is what these benchmarks are after." - I don't get what this mean. Like they are after it like they want it, or they are after it like they succeed it?

docs/src/experimental_framework.md
```diff
@@ -92,6 +92,42 @@ devbind --device '<pci_addr>' --bind uio_pci_generic
...
+data layout, garbage collection, and background maintenance operations removed,
+the results more directly reflect CPU scheduling, interrupt handling, and I/O
+submission and completion paths.
+
+The figures this produces may therefore exceed those seen in real-world,
```
naddinadja: I wouldn't make a new paragraph here, since you use "therefore", which directly refers to the previous sentence.

docs/src/experimental_framework.md
```diff
@@ -151,6 +151,13 @@ Results are rendered as an interactive HTML page at
 ``cijoe-output/artifacts/benchmark-results.html``, which allows results
 from different parameterizations to be compared.
 
+The two workflows that quote an ``nvbandwidth`` reference, ``bench_pcie.yaml``
+and ``bench_cuda_iosize.yaml``, measure it as a step of their own run rather
+than carrying a value fixed across experiments. The measurement varies by well
+under a percent between runs on the same host, so the two experiments quote
+figures that differ slightly in the last digit; each is the reference belonging
+to the run it is reported beside.
```
naddinadja: I wouldn't keep the last sentence ("; each is the reference ..."). The paragraph makes sense without it.

docs/src/experiments/pcie_saturation.md
```diff
@@ -67,6 +69,15 @@ excluding the process startup and teardown that also fall inside the monitoring
 window. **xnvmeperf** reports payload bytes per second based on completed I/O
 operations and their requested sizes.
 
+A sample counts as transferring when PCIe receive traffic exceeds 100 MB/s or
+the graphics engine is more than 1% active. Against a saturated link that floor
+is permissive. Samples taken while the link is ramping up or draining clear it
```
naddinadja: What do you mean that the floor is "permissive"?

docs/src/experiments/pcie_saturation.md
```diff
@@ -67,6 +69,15 @@ excluding the process startup and teardown that also fall inside the monitoring
...
+and are kept, which pulls the mean below the rate the run actually sustained and
+leaves the extremes unrepresentative. At 8 KiB the retained maximum even exceeds
```
naddinadja: "or draining clear it and", i have no idea what this means
and in the end of the sentence: do you mean "leaves the extremes unrepresented"?

docs/src/experiments/pcie_saturation.md
```diff
@@ -130,11 +140,28 @@ cases, exceeding the ``nvbandwidth`` reference of 53.3 GB/s and
...
 with a single CPU thread and only four NVMe devices, demonstrating that the
 uPCIe-cuda path requires minimal CPU and device resources to fully utilize the
 PCIe link at larger I/O sizes.
 
+### GPU Involvement and Run Validity
+
+SM_ACTIVE (1002) is 0.0% at every I/O size, in both the mean and the maximum
+retained sample. Restricted to the transferring samples, that reading is
+positive evidence rather than an artifact of averaging over an idle window: no
+GPU kernel ran at any point while the devices were moving data, confirming that
```
naddinadja: Is "no GPU kernel ran" the right way to express that they were not active? I don't really like "ran"

naddinadja: What about saying they were idle, because I assume they are still "running" (turned on), just not active?

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -104,42 +111,159 @@ The minimum queue depth required to saturate the device varies by ``nqueues``:
...
 
-``nqueues=4`` with ``qdepth=64`` represents the most practical configuration:
-it saturates the device at 4096 total threads, matching ``nqueues=2`` at
-``qdepth=128``, while requiring only half the per-queue depth, halving the
-number of commands in flight per queue.
+On thread count alone, ``nqueues=4`` with ``qdepth=64`` is as economical as
+``nqueues=2`` with ``qdepth=128``: both saturate the devices at 4096 total
+threads, and the former needs only half the per-queue depth, halving the number
+of commands in flight per queue. Thread count is not the only cost, however. The
+activity results below show that the GPU compute the polling kernel occupies is
```
naddinadja: Insert a which or restructure slightly: "the activity results below show that the GPU compute which the polling kernel occupies" OR "he activity results below show that the GPU compute occupied by the polling kernel"

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -104,42 +111,159 @@ The minimum queue depth required to saturate the device varies by ``nqueues``:
...
+set by the queue count, which separates these two configurations decisively.
 
 ### GPU Compute Cost of the Polling Kernel
 
+The IOPS above are one result axis; what the GPU pays to reach them is the
```
naddinadja: "Result axis" sounds kinda weird to me. Would it be better to start the results section with something like "we have two results: the iops and the cost from the GPU", and then have a header for each of them? Because right now there's just "Results: ... IOPS" and then this header

naddinadja: this would also remove the use of semicolon in this sentence, which i think would be nice :)

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -104,42 +111,159 @@ The minimum queue depth required to saturate the device varies by ``nqueues``:
...
 
-The activity plot holds ``nqueues`` at 1 so queue depth is the only variable:
-the total thread count (``qdepth × nqueues × ndevs``) grows from 16 threads at
-``qdepth=1`` to 8192 at ``qdepth=512`` along the x-axis, and fields 1002/1003
-show how much GPU compute capacity that polling footprint costs — the second
-result axis to weigh against the IOPS gained by deeper queues.
+The two fields describe different scarcities. SM activity is the ratio of cycles
+a multiprocessor has at least one warp assigned, averaged over all of them, so
+for a kernel that stays resident it reports breadth: how much of the device the
+poller is present on. Warp slot occupancy is the ratio of resident warps to the
```
naddinadja: The second sentence "SM activity ...": I don't get this one. SM activity is the ratio of which two things? Should I read it as "SM activity is the ratio of (cycles a multiprocessor has at least one warp assigned, averaged) over (all of them)" ? But I still don't think that makes sense - and when writing "all of them", what is "them"?

But then what happens in the follow up? "so for a kernel that stays resident it reports breadth". Is "it" referring to "SM activity?". 

Idk, this whole sentence is really complex to read tbh

naddinadja: Ok we talked about this in person, I suggested shortening it to this: "SM activity (link dcgm) for a kernel that stays resident reports how much of the device the poller is present on"
You don't like this wording precisely, but I'll let you workshop it for a bit haha

naddinadja: Definition here: https://docs.nvidia.com/datacenter/dcgm/latest/user-guide/feature-overview.html

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -104,42 +111,159 @@ The minimum queue depth required to saturate the device varies by ``nqueues``:
...
+| ``nqueues`` | Blocks (``nqueues × ndevs``) | SM active (depth-sweep mean) | IOPS at ``qdepth=128`` |
+| ----------- | --------------------------- | ---------------------------- | ---------------------- |
+| 1           | 16                          | 13.3%                        | 42.4 M                 |
+| 2           | 32                          | 26.8%                        | 61.5 M                 |
+| 4           | 64                          | 53.2%                        | 61.4 M                 |
+| 8           | 128                         | 95.7%                        | 61.4 M                 |
+| 16          | 256                         | 95.9%                        | 61.4 M                 |
+
+The activity column is the mean of each line over all ten queue depths: no depth
+departs from its queue count's value by more than 1.7 percentage points. The
```
naddinadja: "departs" is kinda weird, can we say "differ"?

naddinadja: Also, when you say "its queue count's value", you mean the averaged value, right? Would it be better to say "The SM active column is the mean of each line over all ten queue depths. For every number of queues, the activity did not differ by more than 1.7 percentage points from the average."

docs/src/experiments/device_initiated_qdepth.md
```diff
@@ -104,42 +111,159 @@ The minimum queue depth required to saturate the device varies by ``nqueues``:
...
+would be, and the proportion holds through ``nqueues=4``. At ``nqueues=8`` the
+128 blocks outnumber the 114 multiprocessors, so activity saturates near 96% and
+the further doubling to 256 blocks adds nothing. The knee therefore falls where
+``nqueues × ndevs`` passes the multiprocessor count, not at any queue count in
+particular: with fewer multiprocessors on the GPU, or with more devices
+attached, fewer queues would reach it.
+
+#### The Cost of Queue Depth
+
+Warp slot occupancy is the other axis, and both of its regimes follow from one
```
naddinadja: What are we referring to here when saying the "other" axis? if it's the axes i talked about earlier, it doesn't really make sense, since the "other" axis was already defined there, but I don't find any other mentions of axes.

naddinadja (changes requested):
Okay the code portion looks great, but I have a lot of comments for the text. In general, it's difficult to read, so it took a while. In general, I think it's important that we're "correct" in the way we express ourselves, but it's also very important that we don't do it too complex, if we want our research to be digestable for outsiders. :)

docs/src/experimental_framework.md
karlowich: Dropped that sentence entirely.

docs/src/experimental_framework.md
karlowich: This paragraph referred to the opening statements of the subsection. I have dropped both this paragraph and what it referred to.

docs/src/experimental_framework.md
karlowich: This has been rewritten.

docs/src/experimental_framework.md
karlowich: Agreed

karlowich: Actually, I moved this section to the first experiment using nvbandwidth and rewrote it a bit.

docs/src/experiments/pcie_saturation.md
karlowich: I have rewritten this.

docs/src/experiments/pcie_saturation.md
karlowich: I have rewritten this.

docs/src/experiments/pcie_saturation.md
karlowich: I have rewritten this

karlowich:
@naddinadja I paused work on this for now. I pushed the fixes for the first couple of comments.

karlowich:
@naddinadja I have substantially rewritten the text and changed the commit ordering.
The python changes haven't been touched only the docs is different

safl (approved):
I have not taken a close look at this, but i see many useful topics covered so if somebody else puts in the review effort then i look forward to seeing this go in. So, from my side this, this looks excellent.

naddinadja (approved):
looks good! :)

PR #82, opened 2026-08-25
Replace udmabuf-import patch with dma-buf import module

karlowich:
This moves the provisioning / docs away from building a custom kernel to a stock kernel with a out-of-tree kernel module.


naddinadja (approved):
lgtm :)

PR #83, opened 2026-08-26
fix(setup): report the dma-buf importer's own version 

karlowich:
The package version sat buried in the deb filename, so the docs table printed the uPCIe release, v0.6.0, as the version of the importer, which is at 0.2.0.

The config now carries the package version and the task composes the asset name around it. The install check greps dkms status for it, so a mismatch fails the workflow.


naddinadja (approved):
lgtm :)

PR #84, opened 2026-09-10
feat(setup): install the uPCIe modules as one package

karlowich:
uPCIe now ships the dma-buf importer and the IOMMU physical-address mapper as a single package that loads both modules and installs both UAPI headers by itself.

Provisioning fetches one asset and checks both modules. One config value names the release tag and the deb, since the package version tracks uPCIe.


safl (approved):
lgtm

PR #85, opened 2026-09-14
CI: run provisioning and linting in GitHub Actions

karlowich:
This tries to catch as many problems as possible without a GPU.
It lints all the workflows by running "dry-running" them and it runs the provisioning.


.github/scripts/cijoe_lint.py
```diff
@@ -0,0 +1,98 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+Lint the cijoe workflows without a target
+=========================================
+
+Loads each workflow as 'cijoe' does before running it: scripts resolve,
+templates render against the configs, and each step's 'with' parses.
+'cijoe --integrity-check' ignores the configs. Without '-c', every
```
safl: `cijoe --integrity-check` does load the configs: `cli_integrity_check` builds the combined toml and calls `dict_substitute(workflow_dict, config.options)`. It passes the options dict as the context itself, with no `config`, `local` or `resources` key, so `cijoe --integrity-check -w tasks/bench_cuda_iosize.yaml -c configs/transport.toml -c configs/aisio.toml` reports `'config' is undefined` and `'local' is undefined`. That, together with the `script.load()` and `with` parsing it never does, is what this script adds over it.

scripts/qemu_guest_start_nvme.py
```diff
@@ -0,0 +1,98 @@
...
+    ]
+
+    return [drive], args
+
+
+def main(args, cijoe):
+    drive_size = "8G"
+    guest_name = args.guest_name or cijoe.getconf("qemu.default_guest")
+    if not guest_name:
+        log.error("missing config value(qemu.guest_name)")
```
safl: `qemu.guest_name` is not a config key. Line 74 reads `qemu.default_guest`, and `guest_name` is the `--guest_name` argument.

scripts/stage_nosi_guest.py
```diff
@@ -0,0 +1,66 @@
...
+    url = cijoe.getconf("guest_image.url", None)
+    if not url:
+        log.info("config has no [guest_image].url; nothing to stage")
+        return 0
+
+    from withcache import oras
+
+    grow = cijoe.getconf("guest_image.grow", "8G")
+    image = cijoe.getconf("qemu.default_systemimage", None)
+    dst = Path(cijoe.getconf(f"system-imaging.images.{image}.disk", None)["path"])
```
safl: `getconf` returns its default, `None`, when `qemu.default_systemimage` is unset or the image carries no `disk` table, so this subscript raises `TypeError` instead of reporting. `guest_image.url` on line 28 is guarded; this wants the same.

scripts/stage_nosi_guest.py
```diff
@@ -0,0 +1,66 @@
...
+    dst.parent.mkdir(parents=True, exist_ok=True)
+
+    digest = oras.parse_ref(url).digest
+    stamp = Path(f"{dst}.digest")
+    if digest and dst.exists() and stamp.exists() and stamp.read_text().strip() == digest:
```
safl: The stamp holds the digest alone, so changing `[guest_image].grow` leaves the qcow2 at its previous size: the `qemu-img resize` on line 59 is skipped along with the pull. Record `grow` alongside the digest at line 65 and compare both here.

safl (commented):
In good shape, and nothing here blocks merging.

The four inline comments are all small and self-contained: a docstring claim about `--integrity-check` that does not match what it does, an error message naming a config key that does not exist, an unguarded subscript on a `getconf` that returns None, and a staging cache keyed on the digest alone.

For a follow-up rather than here: `verify.yml` sets no `permissions:`, although `pages.yml` does, and these jobs run a privileged container that pulls and executes a remote image. The action pins are mixed, `checkout@v7.0.0` and `upload-artifact@v7.0.1` to the patch against `setup-python@v5`, and `@v4` in `pages.yml`. `timeout-minutes: 300` sits ten times over the observed 31 minutes. `cijoe_lint.py:45` passes an empty `argparse.Namespace()`, which disables the reserved-key check in `Workflow.dict_lint`, harmless while no script declares an argument named after a cijoe option. The lint job installs cijoe unpinned from PyPI while provision uses whatever the container image carries. `root_unlock.py` sets root's password to the constant `"root"` and enables root password login on a script marked `Retargetable: True`, which is worth confining to a guest or moving into config. Once #86 lands, this branch also needs a `pre-commit run --all-files` pass, since `ruff format` wants four of the five new Python files and yamlfmt re-indents `verify.yml` with the anchors intact, and `.github/scripts` should go into `ruff.toml`'s `src` at the same time.

Run: `.github/scripts/cijoe_lint.py` against the repository configs, reporting ok for all 17 workflows; `cijoe --integrity-check` on `tasks/bench_cuda_iosize.yaml`, which fails as described inline; `pre-commit run --from-ref origin/main --to-ref` this head, which passes; and `ruff format --check` with #86's config over the new files. No QEMU guest was booted, so `guest_start.yaml`, `stage_nosi_guest.py`, `root_unlock.py` and `qemu_guest_start_nvme.py` were read rather than run, and the `systemd-run --on-active=2 systemctl reboot` change was not exercised against any target.

Reviewed with a coding agent. What was built or run is stated above; anything not stated was read rather than exercised. Push back where it is wrong.

Assisted-by: Claude Code:claude-opus-5

.github/scripts/cijoe_lint.py
karlowich: Corrected the docstring

scripts/qemu_guest_start_nvme.py
karlowich: fixed

scripts/stage_nosi_guest.py
karlowich: This has been guarded now

scripts/stage_nosi_guest.py
karlowich: fixed

karlowich:
@safl Addressed all your feedback.
I decided to put your suggestions for follow-up work into the existing commits as this PR is still small.
Now the following changes have been made:
- `verify.yml` now has `contents: read`, like `format.yml`.
- The lint job now runs in the same nosi container as the provision jobs, using that container's cijoe Python
- Provision's timeout-minutes goes from 300 to 90
- `cijoe_lint.py` now builds the options with cijoe's own `parse_args()`
- `root_unlock.py` now marked `Retargetable: False`. The docstring says it sets a fixed root password and turns on password login, so it's only for throwaway guests.

safl (approved):
lgtm!

PR #86, opened 2026-09-15
CI: run format check

karlowich:
This will format python and yaml on every PR.
Additionally, it adjusts the shellcheck to only fail on errors.


safl (approved):
Great, let's get this in first such that the other open PRs can get their formating fixed up.

